PR #10, opened 2020-02-19
fixing ipywidgets

choldgraf:
This is an attempt at fixing ipywidgets following the suggestions of @akhmerov...for some reason it still doesn't work though :-/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,6 @@ jobs:
           keys:
             - cache-pip
 
-      - run: pip install --user -r requirements.txt
       - run: pip install --user -e .[sphinx,rtd]
       - save_cache:
           key: cache-pip
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -44,7 +44,7 @@ exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "pandas_sphinx_theme"
+html_theme = "alabaster"
 html_logo = "_static/logo.png"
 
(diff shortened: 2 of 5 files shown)


akhmerov:
Still [this](https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/execute.py#L952-L959) to add?

choldgraf:
@akhmerov i believe that is here: https://github.com/ExecutableBookProject/sphinx-notebook/blob/master/sphinx_notebook/__init__.py#L83

akhmerov:
Uhhh, maybe [this](https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/execute.py#L844-L845) then? :)

akhmerov:
Or otherwise did you verify that the notebook that you're converting has widget state metadata stored?

choldgraf:
I believe it does -if I open it up in jupyterlab then the widgets pop back up. I added in the javascript code from the function you mentioned - that definitely should be there either way

choldgraf:
ugh - OK I think I figured out what the problem is - bootstrap (loaded by the theme we're using) seems to be clashing with RequireJS (required by ipywidgets). It's not related to this codebase at all :-P

I opened up https://github.com/jupyter-widgets/ipywidgets/issues/2787 and https://github.com/pandas-dev/pydata-bootstrap-sphinx-theme/issues/93 to try and spot-check this.

@akhmerov do you know if anybody has gotten jupyter-sphinx to run on a bootstrap-based site?

choldgraf:
Latest commit should get this working by switching themes to Alabaster, which doesn't have this problem. These docs aren't complex enough yet to warrant a theme like pydata anyway, so we can test on alabaster and spot-check the pydata them problem in another issue

PR #20, opened 2020-02-21
not using document, also fixing widget state

choldgraf:
closes https://github.com/ExecutableBookProject/sphinx-notebook/issues/18

booyah figured it out, thanks @chrisjsewell for the help


akhmerov:
Interesting, so including the widget state node wasn't necessary after all?

choldgraf:
@akhmerov oh no it was, I am just an idiot and was including it twice :-)

PR #28, opened 2020-02-22
parsing notebook-level metadata as yaml

choldgraf:
Doesn't work yet - but the idea is that this should behave the same way that the myst parser parses front-matter metadata. It probably isn't working because the yaml.dump isn't treating the metadata as a dict, but instead as a notebooknode.


chrisjsewell:
FYI I found out recently that you can create specific [draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which this would be. Unfortunately you can't yet convert an existing PR to a draft one.

chrisjsewell:
looks good though 👍

choldgraf:
Any reason that I would be getting this error:

```
Exception occurred:
  File "/home/choldgraf/anaconda/envs/dev/lib/python3.7/site-packages/docutils/nodes.py", line 208, in _all_traverse
    result.extend(child._all_traverse())
AttributeError: 'FrontMatter' object has no attribute '_all_traverse'
```

chrisjsewell:
Yeh because you are trying to parse Markdown AST to docutils lol.
You need to replicate what [DocutilsRender.render_front_matter](https://github.com/ExecutableBookProject/myst_parser/blob/7cc2c92eb6fc1678b4faeefb7b3f6a204473dc1a/myst_parser/docutils_renderer.py#L97) does.

choldgraf:
ah damnit I keep losing track of all the AST -> AST conversions here, 👍

chrisjsewell:
Noticed a typo in that docstring actually; it should say 'occurs at the top of the document`.

I guess ideally the main code from `render_front_matter` should be made standalone, so that you can call it to get the `docinfo` node.

choldgraf:
Maybe it could call a little helper function inside that I can re-use? Otherwise I will probably just re-write the internals to use here.

That said, one thing I noticed, from https://github.com/ExecutableBookProject/myst_parser/blob/7cc2c92eb6fc1678b4faeefb7b3f6a204473dc1a/myst_parser/docutils_renderer.py#L124 it makes me feel like it is not possible to have nested metadata in front-matter YAML. Is that true?

e.g. I thought we could do things like

```
---
key:
  key2: val2
  key3: val3
---
```
but maybe it's not supported currently?

chrisjsewell:
Maybe you want to make a PR to `myst_parser` to split `render_front_matter` into an `@staticmethod` part that creates the `docinfo` from a parsed dict. Then you could call `SphinxRenderer.create_docinfo(dict)` 😁

choldgraf:
I can give it a shot but first I gotta sleep :-)

chrisjsewell:
> possible to have nested metadata in front-matter YAML

Its possible in the Markdown AST, but the docutils AST requires that fields be string values, so you could, but only if you JSON serialized them first.

chrisjsewell:
[here](https://github.com/ExecutableBookProject/myst_parser/blob/7cc2c92eb6fc1678b4faeefb7b3f6a204473dc1a/myst_parser/docutils_renderer.py#L125) I chuck away anything not a str/float/int, but probably should just serialize everything to strings.

choldgraf:
OK I gave a shot at adding a convenience func to myst_parser and opened a PR there...this won't run until if/when that is merged.

Issue #30, opened 2020-02-23
Section headings in markdown render nodes outside of the InputCell nodes


choldgraf:
I just noticed that we have a bug in how the rendered docutils nodes are put (or not) inside of the `CellInputNode` that should contain them. I *think* I know what the problem is.

For each markdown cell, we create a `CellInputNode` and set it as the current node. Then we render the markdown content inside. What we want is for whatever nodes are spit out to be put *inside* the `CellInputNode` (as children). This usually works for most markdown. However, if there are section headers, or other syntax that should trigger a new *section*, then the render function creates a *new section* and starts rendering *outside* of the `CellInputNode`. I think this is where that happens for headers:

https://github.com/ExecutableBookProject/myst_parser/blob/develop/myst_parser/docutils_renderer.py#L259

@chrisjsewell can you think of a way that we can ensure the renderer is always placing objects inside the `CellinputNode`?

choldgraf:
One option that would be fairly hacky would be to, at the end of the "render" operation, check if there are any non `CellNode` nodes in the doctree. And if so, assume that they should be appended to the children of the latest `CellInputNote`...

chrisjsewell:
Hmm thats a tricky one. It feels like you have two immiscible hierarchies here; sections and markdown cells, since both can encapsulate each other. Do you actually need to wrap the markdown cells in `CellInputNode`, what would it be used for?  Would it be easier to do what the `SphinxRenderer` now does and add a 'classed' comment node, to denote a break between two markdown cells, as in [`render_block_break`](https://github.com/ExecutableBookProject/myst_parser/blob/e463e0afea75b7d68dbe38dc8f0c40da834e9a0d/myst_parser/docutils_renderer.py#L184)

choldgraf:
Well,  the main trick is that `jupyter-book` often wants to know about a "block" of markdown to do things to it (e.g. "look for the cell tag `div.cell.tag_hide_input` and when found, hide the input of that cell"). So keeping things together with their metadata is important to Jupyter Book at least. The other nice part about this from a CSS perspective is that you can use a single selector (e.g. `div.cell`) and know that everything on the page will behave the same way at the top level.

1. Special-case the page `title` field and have some combination of "check ntbk-level metadata, and if no `title` field is there, find the first markdown heading and treat it as the "title" for the document". Then the title node gets added before any subsequent cell nodes are added.
[...]

PR #31, opened 2020-02-23
reworking cell and markdown relationship in parser

choldgraf:
Maybe this could be a solution to #30 - it does a lot of manual re-arranging such that I'd bet I'm breaking something though...

~This *sort of* works, but it also then throws a bunch of warnings for pages, because there is no top-level "title" for the page because they're all wrapped in `CellNode` blocks~

## Updated PR

The updated version of this PR takes a different approach, that I think is better. Rather than forcing everything to be in a top-level `cell` div, this PR now parses the notebook more "naturally". It assumes that code blocks are not necessarily in totally different containers, but instead will flow with the rest of the AST as if they were "literal" blocks.

This means that the AST of the document now looks much more like a regular AST from parsing a text file. We still use the `CellNode` containers to wrap the code cells in `<div>` elements for tagging, and the "hide input" behavior is still there.

For markdown cells, I added a dependency on [`sphinx-togglebutton`](https://github.com/executablebookproject/sphinx-togglebutton) and added documentation for how they can use directives to make toggle-able markdown, e.g., with:

````
```{note}
:class: toggle
My note content
```
````
We might want to change up that API in https://github.com/executablebookproject/sphinx-togglebutton but that's where it is currently.

I also added a bit of logic that will check if *hide_input* is present in a markdown cell and, if so, will also add a container to wrap in a div for the toggle button to find.

@chrisjsewell what do you think? does this make more sense?

Here's how it looks: https://48-241268229-gh.circle-artifacts.com/0/html/notebooks.html


choldgraf:
The latest push implements some of what we want, I think. It adds a custom HTML5Translator class that adds `<div>` elements before and after comments with specific tags.

## Some problems w/ this PR

* For some reason, the second `</div>` tag is closing the `<body>` element of the page. I wonder if this is actually uncovering a bug in the HTML of the parser or theme somewhere...

## A potential alternative to this PR

Another option we could explore is to *only wrap markdown cells in divs if they have tags*. Then we could use the same "container" approach used originally in this repo, with two caveats

* the first markdown cell couldn't be used w/ containers (because it'd be used to define the page title etc)
* the CSS and JS rules of the page would have two different kinds of content to worry about, "things in cell divs", and "other things". However, I believe that Sphinx will always output a `<div class="section"` so that might be usable as a selector as well.

choldgraf:
OK I'm gonna merge this because I'm _pretty_ sure that @chrisjsewell would agree w/ the changes :-)

chrisjsewell:
Without any tests, who knows if it works❓❗️ 😆

PR #40, opened 2020-03-03
Fix Install instructions

chrisjsewell:


docs/index.md
```diff
@@ -21,12 +21,12 @@ To install `myst-nb`, do the following:
...
   `myst_parser`:
   ```
-  pip install -e "git+https://github.com/ExecutableBookProject/myst_parser.git#egg=myst_parser[sphinx]
+  pip install -e "git+https://github.com/ExecutableBookProject/myst_parser.git#egg=myst_parser[sphinx]"
   ```
 
   `myst-nb`:
   ```
-  pip install -e "git+https://github.com/executablebookproject/sphinx-notebook.git#egg=sphinx-notebook[sphinx]
+  pip install -e "git+https://github.com/executablebookproject/sphinx-notebook.git#egg=sphinx-notebook[sphinx]"
```
chrisjsewell: What is the `sphinx` extra actually for? I'm guessing for RTD builds? If so would it be better to rename it to `rtd`, since it has a tonne of dependency that I'm sure people don't actually need to use MyST-NB.

chrisjsewell: (for `myst-parser` the `sphinx` extra is only for packages that you actually need to use it as a sphinx extension)

choldgraf: my guess is it's copypasta from the parser :-) maybe we should make this "docs" so it's clear that these are the packages needed to build the docs from scratch?

choldgraf: and either way, I agree we should remove the extra requirement field from the basic install instructions

chrisjsewell: You already had an `rtd` extra as well! So I merged the two in 20b544b

choldgraf:
sweet! LGTM and the tests are passing*!












* (because there are no tests!)

PR #41, opened 2020-03-03
using document and adding a more helpful error

choldgraf:
This uses `Document` instead of tokenize, per @chrisjsewell 's latest comments about improvements to that class.

It also adds a simple try/except block to capture any errors associated with parsing the notebook and display a message that includes the notebook cell and path. @chrisjsewell do you think this is a reasonable way to handle this? Know of a better way to get the error message to display? I know it's not idea because the errors won't propagate past this step in Sphinx, but as a stopgap is it the right improvement?

### DIFF
--- a/myst_nb/parser.py
+++ b/myst_nb/parser.py
@@ -1,8 +1,9 @@
 from docutils import nodes
 import nbformat as nbf
+from pathlib import Path
 
 from myst_parser.docutils_renderer import SphinxRenderer, dict_to_docinfo
-from myst_parser.block_tokens import tokenize
+from myst_parser.block_tokens import Document
 from myst_parser.sphinx_parser import MystParser
 from jupyter_sphinx.execute import get_widgets, contains_widgets, JupyterWidgetStateNode
 
@@ -19,6 +20,7 @@ class NotebookParser(MystParser):
     config_section_dependencies = ("parsers",)
 
     def parse(self, inputstring, document):
+        self.reporter = document.reporter
         self.config = self.default_config.copy()
         try:
             new_cfg = document.settings.env.config.myst_config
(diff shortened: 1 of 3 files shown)


myst_nb/parser.py
```diff
@@ -40,53 +41,62 @@ def parse(self, inputstring, document):
...
+                    doc2path = document.settings.env.doc2path
+                    path_page = Path(doc2path(document.settings.env.docname))
+                    srcdir = Path(document.settings.env.srcdir)
+                    relpath = path_page.relative_to(srcdir)
+                    raise Exception(
+                        f"Error parsing notebook {relpath}, cell #{ii+1}: {exc}"
+                    )
+
+
+def _render_cell(cell, renderer):
```
choldgraf: none of the logic here is different (other than using `Document`), it's just moved into a function to make it easier to human-parse

myst_nb/parser.py
```diff
@@ -40,53 +41,62 @@ def parse(self, inputstring, document):
...
-                    outputs = CellOutputBundleNode(cell["outputs"])
-                    cell_output += outputs
+                try:
+                    _render_cell(cell, renderer)
+                except Exception as exc:
+                    doc2path = document.settings.env.doc2path
+                    path_page = Path(doc2path(document.settings.env.docname))
+                    srcdir = Path(document.settings.env.srcdir)
+                    relpath = path_page.relative_to(srcdir)
+                    raise Exception(
```
choldgraf: Is there a better way of doing this? Perhaps with the Sphinx logging instance?

choldgraf:
updated this PR with (I think) a better error reporting mechanism (now using `self.reporter`)

PR #52, opened 2020-03-07
Add transform tests

chrisjsewell:
Immediately I found a bug lol; that you were looking for images then treating them all as png. I have added separate logic for JPEG and a 'TODO' for SVG

**NOTE**: I don't know if any of the test outputs are *actually* correct. This just tests that the parser/transform completes without exception, and records the outputs, so that any changes to them will be recorded if the code changes.

As I've documented in MyST-Parser, this uses pytest-regressions, to store and diff against the test outputs. To refresh `pytest --force-regen`


choldgraf:
Ah fantastic, thanks for starting this process. There will definitely be more bugs so getting tests in will be super helpful

PR #55, opened 2020-03-09
FEAT: Jupyter-cache integration

AakashGfude:
Simple implementation of integration of `jupyter-cache` in `myst-nb's` parser. 

Current implementation

- [x]  gets a local db instance, else creates it (implementation in `jupyter-cache`)
- [x]  stages the notebook
- [x]  executes the notebook
- [x]  merges the executed notebook with the original one and forwards it for parsing.

Apart from making the present implementation more robust, I feel a few functionalities need to be present in MyST-NB :- 

- **Having the functionality to use the `clear_cache` function of `jupyter_cache`**. It might be necessary if someone wants to re-execute the notebooks without actually changing the code of any of them
- **parallel execution of notebooks**. Since, the parser function takes one notebook at a time, implementing parallel execution in `jupyter-cache` level won't be enough, and parallel execution cannot be implemented in parser I think,  because it will create a new instance of multiprocessing library for each notebook. 

My suggestion for both of the above functionalities is to implement them in a sphinx builder.
 
Please provide your suggestions @choldgraf @chrisjsewell @mmcky, I will meanwhile continue the implementation to handle as many cases as possible.


AakashGfude:
Also,  we should have a proper traceback for debugging execution errors, primarily a report file to store all the tracebacks.
At present, we have an error message indicating, the failed notebook.
 
image

chrisjsewell:
> Also, we should have a proper traceback for debugging execution errors, primarily a report file to store all the tracebacks.
> At present, we have an error message indicating, the failed notebook.
> 

Note that all tracebacks (and eventually failed notebooks) are stored in the cache, see: https://github.com/ExecutableBookProject/jupyter-cache#staging-notebooks-for-execution

AakashGfude:
> > Also, we should have a proper traceback for debugging execution errors, primarily a report file to store all the tracebacks.
> > At present, we have an error message indicating, the failed notebook.
> 
> Note that all tracebacks (and eventually failed notebooks) are stored in the cache, see: https://github.com/ExecutableBookProject/jupyter-cache#staging-notebooks-for-execution

Great, thanks. So, I should recover the cache information using the pk of the notebook. I will test this out tomorrow morning.

chrisjsewell:
> Great, thanks. So, I should recover the cache information using the pk of the notebook. I will test this out tomorrow morning.

Yes, note that it is stored on the staged record, not a cache record, because obviously since the notebook failed, it is not cached and there is no cache record created

chrisjsewell:
> parallel execution of notebooks. Since, the parser function takes one notebook at a time, implementing parallel execution in jupyter-cache level won't be enough, and parallel execution cannot be implemented in parser I think, because it will create a new instance of multiprocessing library for each notebook.

I mentioned this in an issue before (can't remember which one!), but I feel that this should be handled should be handled around step (3) of the build process below, whereby you stage all notebooks, then call execute. Then, in the parse step (7), you will just attempt to retrieve a cached notebook or, on `KeyError`, the trackback from the staged record.

```
1. event.config-inited(app,config)
2. event.builder-inited(app)
3. event.env-get-outdated(app, env, added, changed, removed)
4. event.env-before-read-docs(app, env, docnames)

for docname in docnames:
    5.  event.env-purge-doc(app, env, docname)
    if doc changed and not removed:
        6. source-read(app, docname, source)
        7. run parser: text -> docutils.document
        8. apply transforms (by priority): docutils.document -> docutils.document
        9. event.doctree-read(app, doctree)

10. event.env-updated(app, env)
11. event.env-check-consistency(app, env)

for docname in docnames:
    12. apply post-transforms (by priority): docutils.document -> docutils.document
    13. event.doctree-resolved(app, doctree, docname)

14. call builder

15. event.build-finished(app, exception)
```

chrisjsewell:
FYI, before this is merged, make sure to add some tests, using the notebooks in https://github.com/ExecutableBookProject/MyST-NB/tree/master/tests/notebooks (for an example of a full sphinx integration test, see https://github.com/ExecutableBookProject/MyST-Parser/blob/master/tests/test_sphinx/test_sphinx_builds.py)

choldgraf:
thanks for the first pass! a few quick thoughts from me:

* I may make some pushes to this branch today to add some extra functionality
* I'm a bit hesitant for MyST-NB to do much in the way of fancy execution. I think that as a first step, we should focus on "simple" execution (e.g. add two configuration variables: one for "execute" one for "cache". If the first is triggered, all ipynb files will be executed at build. If the second is triggered, all ipynb files will be executed and cached, and subsequent builds will use the cache. I think we should discuss whether this tool should be used for things like large-scale parallel execution, or whether that should be instead built into `jupyter-cache` and manually run from the CLI for larger jobs.
* Agreed people need to be able to clear the cache...could we assume that users can just delete the `.jupyter_cache` folder in that case? Is there a way to do this outside of a CLI? (since myst-nb has no CLI...perhaps we can just recommend the jupyter-cache CLI once it solidifies)

Another thing to remember - users of Jupyter Book should need to know nothing about MyST-NB, we could also make some of these improvements (e.g. clearing cache and parallel execution) in the `jupyter-book` CLI. The MyST-NB repo might be used directly by software developers in their Sphinx builds, so we can make more advanced assumptions about their comfort with coding in general.

choldgraf:
I made a PR to your branch @AakashGfude - I had started some work on this earlier, and that PR is to help give an idea where I was going, and give you some code to work with if it's helpful 👍

AakashGfude:
> * I may make some pushes to this branch today to add some extra functionality
Thank you so much. I will have a look at the PR.

> * Agreed people need to be able to clear the cache...could we assume that users can just delete the `.jupyter_cache` folder in that case? Is there a way to do this outside of a CLI? (since myst-nb has no CLI...perhaps we can just recommend the jupyter-cache CLI once it solidifies)

I assumed the user will have no knowledge about the internals of caching and DB being used. And imagined that the command to clear cache would be done in the background, when something like `make clean` or `make clean-cache` is run And, also since `.jupyter_cache` folder is not visible in the file explorer. But I guess, most if not all the users might be using a text editor anyway so visibility might not be an issue. I just wonder how intuitive it will be to delete the folder.  

> Another thing to remember - users of Jupyter Book should need to know nothing about MyST-NB, we could also make some of these improvements (e.g. clearing cache and parallel execution) in the `jupyter-book` CLI. The MyST-NB repo might be used directly by software developers in their Sphinx builds, so we can make more advanced assumptions about their comfort with coding in general.

This is a nice idea as well. We will need to decide which layers should have what functionalities. I am not entirely clear about the functionalities `jupyter-book` is supposed to have. missed the discussions. will get a clearer idea.

> * I'm a bit hesitant for MyST-NB to do much in the way of fancy execution. I think that as a first step, we should focus on "simple" execution (e.g. add two configuration variables: one for "execute" one for "cache". If the first is triggered, all ipynb files will be executed at build. If the second is triggered, all ipynb files will be executed and cached, and subsequent builds will use the cache. I think we should discuss whether this tool should be used for things like large-scale parallel execution, or whether that should be instead built into `jupyter-cache` and manually run from the CLI for larger jobs.

Like you said above, users of Jupyter Book need not know about MyST-NB and internals. So we should avoid situations in which they might have to use the internal complicated tools?

AakashGfude:
> I mentioned this in an issue before (can't remember which one!), but I feel that this should be handled around step (3) of the build process below (by adding an [event callback](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events)), whereby you stage all notebooks, then call execute. Then, in the parse step (7), you will just attempt to retrieve a cached notebook or, on `KeyError`, the trackback from the staged record.

I had also imagined in the same lines. @choldgraf what are your thoughts on this?

choldgraf:
yeah that sounds reasonable to me - it'll be hard to tell without a prototype to play around with though, I'm happy to try something out when it's ready

AakashGfude:
@choldgraf @chrisjsewell have reorganized the code in the last few commits *as a test*, according to the discussion with  @chrisjsewell above regarding parallel execution, where the notebooks are executed and cached during the `env-get-outdated`  event stage. And in the parser, the cached outputs and original notebooks are merged. 

Feel free to scrap this idea, or embrace it.

AakashGfude:
Also @choldgraf , I got this attribute error for genutils after merging with master. Did the merge tamper anything?

image

PS:- I will remove this comment later to prevent cluttering.

chrisjsewell:
Sorry @AakashGfude with 688e4211225b697f781519a3a53cac0497b601fb you'll need to refactor again 😬

myst_nb/cache.py
```diff
@@ -0,0 +1,128 @@
...
+Implements integration of jupyter-cache
+"""
+import os
+import json
+import nbformat as nbf
+from pathlib import Path
+from sphinx.util import logging
+from sphinx.util.osutil import ensuredir
+from jupyter_cache.cache import main as cache
+from jupyter_cache.cli.utils import get_cache
```
chrisjsewell: This has now been moved to just `from jupyter_cache import get_cache`

setup.py
```diff
@@ -47,6 +47,8 @@
         "jupyter_sphinx==0.2.4a1",
         "nbformat",
         "nbconvert",
+        "nbclient",
+        "jupyter-cache",
```
chrisjsewell:
```suggestion
        "jupyter-cache~=0.1",
```

choldgraf:
FYI there is another PR coming down the pipeline that may need another (minor I think) refactor #73

AakashGfude:
@chrisjsewell @choldgraf  because of adding some import statements in notebooks and XML in tests folder, and other changes necessary for error-free execution. There are quite a few diffs between generated and expected. I am just going through those and mending any errors. 
If you guys wanted to have a look, some of them might be more obvious to you guys?

image

Also,  for writing execution tests, I will follow the convention of other tests and use those notebooks. I will probably have functions like test_complex_outputs and test_basic_run, with file_regression between expected ipynb files and produced outputs.

myst_nb/parser.py
```diff
@@ -37,6 +41,25 @@ def parse(self, inputstring, document):
         # de-serialize the notebook
         ntbk = nbf.reads(inputstring, nbf.NO_CONVERT)
 
+        self.reporter = document.reporter
+        self.config = self.default_config.copy()
+
+        try:
```
chrisjsewell: I'm not sure if there is a good reason for all this being in a try/except block

myst_nb/parser.py
```diff
@@ -37,6 +41,25 @@ def parse(self, inputstring, document):
...
+        except AttributeError:
+            pass
+            
+        ## add outputs to notebook from the cache
+        if hasattr(document.settings.env, 'path_cache'):
+            path_cache = document.settings.env.path_cache
+            ntbk = add_notebook_outputs(source_path, ntbk, path_cache, dest_path)
+        else:
+            # If we explicitly did not wish to cache, then just execute the notebook
+            ntbk = execute(ntbk)
```
chrisjsewell: I think there should be a config option to not do any execution. This would also be useful for testing, because then we could keep the original tests, like complex_ouputs, that are more focussed on the output creation, and add new notebooks/tests specific to execution

chrisjsewell:
> There are quite a few diffs between generated and expected. 

As I mentioned in the review above I think you should probably keep the original tests how they are, using a configuration option in the tests to signify that no execution should be run. Then add new tests that are focussed on mainly the execution aspects.

myst_nb/parser.py
AakashGfude: the block was already there, with some variables, I just added mine. Probably, I should move mine away from it.

myst_nb/parser.py
AakashGfude: you are right, I was planning to create it as well. Will create one with default False.

.travis.yml
```diff
@@ -10,8 +10,13 @@ matrix:
         env: TEST_TYPE="pytest"
       - python: 3.7
         env: TEST_TYPE="pre-commit"
+
 install:
-  - pip install --upgrade pip wheel setuptools
+  - wget https://repo.anaconda.com/archive/Anaconda3-2019.10-Linux-x86_64.sh -O anaconda3.sh
```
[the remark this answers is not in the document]
AakashGfude: most of the code there has pandas, matplotlib, sympy and others. I will probably create a new one, with at most one or two important ones and put them in the setup.py then

myst_nb/parser.py
```diff
@@ -50,7 +73,7 @@ def parse(self, inputstring, document):
             logger=SPHINX_LOGGER,
         )
         set_parse_context(parse_context)
-
+
```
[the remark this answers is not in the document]
AakashGfude: can you elaborate more on this?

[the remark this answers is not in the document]
chrisjsewell: I have got round to adding my customary contributing file to this repo, but basically: https://myst-parser.readthedocs.io/en/latest/develop/contributing.html, swapping myst-parser for myst-nb, this runs the flake and black formatting/syntax check before every commit, avoiding things like this

myst_nb/parser.py
chrisjsewell: yeh maybe @choldgraf the less in try/except the better

AakashGfude:
@choldgraf @chrisjsewell , so I was trying to write a simple test case today for this and had a few questions on how to implement this. **Please ignore the coveralls report for now.**
As of now, I have just written a simple test to check if `basic_unrun.ipynb`, when executed using the plugin, produces the same output as `basic_run.ipynb` :- https://github.com/ExecutableBookProject/MyST-NB/blob/95eb0a2909b96494c7a909e3c13824797906097d/tests/test_execute.py. I used `nbdime` instead of  using `file_regression` because it didn't seem well fit for notebooks? 
But, I am new to this and might be missing some obvious points, and before I write a few more tests and make it more organied, wanted to get your expert advice on this one.

myst_nb/parser.py
choldgraf: definitely feel free to clobber try/except in order to do a more rigorous check...I think a lot of this was just getting to MVP's for iteration purposes 👍

choldgraf:
Correct me if I'm wrong, but since all of the notebooks are going to then be turned into docutils XML files, we can keep the same regression check, but make sure that the outputs in the `CellOutputBundleNodes` still look correct. (as opposed to using the `ipynb` files themselves for testing). Is that what @chrisjsewell had in mind?

chrisjsewell:
Yeh I think what you have done is reasonable. At this 'level' of testing, we don't need to be checking XML, just that the notebook has executed. 

You get more fancy, and turn the diff into a pretty printed string, to give better feedback on what has changed;a bit like I do in: https://github.com/ExecutableBookProject/jupyter-cache/blob/95017812af123f0f6dc3127d0f12e9517485978b/jupyter_cache/cache/main.py#L386
but its not strictly necessary.

If you wanted to check lots more notebooks, and have the test's expected outcome "self-generate"/reg-generate, which is what pytest-regressions is good at. You can use `file_regression` with a custom `check_fn`, see: https://pytest-regressions.readthedocs.io/en/latest/api.html#file-regression, something like:

```pytest
import nbformat

def check_nbs(obtained_filename, expected_filename):
    obtained_nb = nbformat.read(obtained_filename)
    expect_nb = nbformat.read(expected_filename)
   diff = diff_notebooks(obtained_nb, expect_nb)
   if diff:
       raise AssertionError(pretty_print_diff(diff)

def test_something(file_regression):
    ...
    file_regression.check(nbformat.writes(nb), check_fn=check_nbs)
```

Also, if you wanted to run execution with more complex, non-deterministic outputs (e.g. matplotlib figures), you would need to think about recursively doing a find/replace on some aspects of the obtained_nb/expect_nb (this is kind of what `pytest-notebook` does, but I don't think you need to actually use that).

choldgraf:
Just a note that my intuition is we should check to make sure the right kinds of outputs are generated, and they they are inserted into the doctree in the correct way, and that the final nodes generated are correct, but we shouldn't need to be ensuring that notebooks reproduce the same results up to 10 decimals, or that image pixels are exactly the same...hopefully nbclient's testing is good enough for this :-)

chrisjsewell:
@AakashGfude you do realise that you've just committed a load of jupyter caches with crazy names lol. It looks like your instantiating the cache with a tuple rather than a string for the path

AakashGfude:
@chrisjsewell  yes, saw that, was a bit impatient to see the coveralls result, my bad. I was fixing it locally and trying to cover all the new code in tests, including the generated XML and ipynb.  Do, you have a local setup for coveralls using APIs or something?

chrisjsewell:
> Do, you have a local setup for coveralls using APIs or something?

You literally just run `pytest -v --cov=myst_nb` lol

AakashGfude:
> > Do, you have a local setup for coveralls using APIs or something?
> 
> You literally just run `pytest -v --cov=myst_nb` lol

Ohh is it haha. Doesn't show anything up in the console. Just runs the pytest tests. I looked at the `.travis.yml` file and there is a command specifically for `coveralls` after success.
image

Ran this command, but it only seems to want to run on TravisCI. I am a bit confused. help me lol.

chrisjsewell:
Thats to send the result to coveralls, but running what I said (assuming you have `pytest-cov` installed, gives:

image

or if you run `pytest -v --cov=myst_nb --cov-report=html` it will output the per line coverage to file

AakashGfude:
Thank you so much @chrisjsewell giving target as `html` successfully creates a report as a web page. 
My console was not showing the above report, even though `pytest-cov` is installed. Will check what is wrong.

chrisjsewell:
If you use ‘—cov-report’ then the console report won’t show

AakashGfude:
Aaah. that's the culprit. I had copied the command from `.travis.yml` directly lol.  Thanks a ton @chrisjsewell

AakashGfude:
Hey @chrisjsewell @choldgraf thank you for your advice above. Following that, I was trying to execute a notebook with a complex non-deterministic output and comparing the generated ipynb files. 

- Was using `np.random.seed(42)` to set a seed for the random numbers. 

- import matplotlib.pyplot as 
  ```
  from ipypublish import nb_setup
  plt = nb_setup.setup_matplotlib(print_errors=True, output=('svg',))
  ```
  because pdf output was always genrating random md5.

- used `rcParams['svg.hashsalt'] = 42` , so that the id values and other random values assigned to svg paramters remain constant.

The tests are passing in my local system everytime, but here it is failing. Maybe `svg.hashsalt` changes the output value between systems? I hope it does not, but the diffs look something like :- 
image

I am sort of out of ideas here to compare these outputs. Let me know what ideas you guys have and how you would approach it. 
Now, I feel like this particular test does not hold as much value as the time demanded by it.  Already wasted a bit of time :sweat_smile:

AakashGfude:
The config variables involved here are:- 

```
"jupyter_execute_notebooks": True,
 "jupyter_notebook_force_run": True, # forces execution even if all outputs are already present in source
 "execution_excludepatterns": ['index'], # skipping execution for some patterns
 "jupyter_cache": True, # False, will use nbclient instead
```

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+from jupyter_cache import get_cache
+from jupyter_cache.executors import load_executor
+
+logger = logging.getLogger(__name__)
+
+
+def execution_cache(app, env, added, changed, removed, path_cache=None):
+    """
+    If cacheing is required, stages and executes the added or modified notebooks,
+    and caches then for further use.
```
choldgraf:
```suggestion
    and caches them for further use.
```

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+    jupyter_cache = False
+    nb_list = added.union(
+        changed
+    )  # all the added and changed notebooks should be operated on.
+    if env.config["jupyter_execute_notebooks"]:
+        jupyter_cache = env.config["jupyter_cache"]
+
+    if jupyter_cache is True:
+        path_cache = path_cache or Path(env.outdir).parent.joinpath(".jupyter_cache")
+        app.env.path_cache = (
```
choldgraf: seems reasonable to me

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+            path_cache  # TODO: is there a better way to make it accessible?
+        )
+
+        stage_and_execute(env, nb_list, path_cache)
+
+    return nb_list  # TODO: can also compare timestamps for inputs outputs
+
+
+def stage_and_execute(env, nb_list, path_cache):
+    pk_list = None
```
choldgraf: could you add a docstring for any user-facing functions? (or make it hidden)

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+
+        cache_base = get_cache(path_cache)
+
+        if "jupyter_notebook_force_run" in env.config:
+            do_run = env.config["jupyter_notebook_force_run"]
+
+        for nb in nb_list:
+            # excludes the file with patterns given in execution_excludepatterns
+            # conf variable from executing, like index.rst
+            exclude_file = [x in nb for x in env.config["execution_excludepatterns"]]
```
choldgraf: can you at least use `xx` instead of `x`? Makes `ctrl-f` search easier IME

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+            if True in exclude_file:
+                continue
+
+            has_outputs = False
+
+            if "/" in nb:  # nb includes the path to notebook
```
choldgraf: doesn't doc2path also operate on files with `/` in the name? (e.g. if the file is in a sub-directory of the docs?

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+
+            if not has_outputs:
+                if pk_list is None:
+                    pk_list = []
+                stage_record = cache_base.stage_notebook_file(source_path)
+                pk_list.append(stage_record.pk)
+            else:
+                if has_outputs:  # directly cache the already executed notebooks
+                    cache_base.cache_notebook_file(source_path, overwrite=True)
+                logger.error(
```
choldgraf: this should be indented, right? Also probably `warning` not `error`?

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+    Add outputs to a NotebookNode by pulling from cache.
+
+    Function to get the database instance. Get the cached output of the notebook
+    and merge it with the original notebook. If there is no cached output,
+    checks if there was error during execution, then saves the traceback to a log file.
+    """
+    # If we have a jupyter_cache, see if there's a cache for this notebook
+    file_path = file_path or env.doc2path(env.docname)
+    dest_path = env.app.outdir
+    reports_dir = dest_path + "/reports"
```
choldgraf: use Pathlib in case somebody is on windows

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+        path_cache = env.path_cache
+
+    if path_cache:
+        cache_base = get_cache(path_cache)
+        db = cache_base.db
+        cache_record = None
+        r_file_path = _relative_file_path(file_path)
+        try:
+            cache_list = NbCacheRecord.records_from_uri(file_path, db)
+            if len(cache_list):
```
choldgraf: I believe this can just be `if cache_list`

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+        else:
+            try:
+                stage_record = cache_base.get_staged_record(file_path)
+            except KeyError:
+                stage_record = None
+            if stage_record and stage_record.traceback:
+                # save the traceback to a log file
+                ensuredir(reports_dir)
+                file_name = r_file_path[
+                    r_file_path.rfind("/") + 1 : r_file_path.rfind(".")
```
choldgraf: should these be `os.sep`s? (maybe pathlib is smart about figuring out paths with `/` on non-*nix systems?)

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+                ]
+                full_path = reports_dir + "/{}.log".format(file_name)
+                with open(full_path, "w") as log_file:
+                    log_file.write(stage_record.traceback)
+                logger.info(
+                    "Execution traceback for {} is saved in {}".format(
+                        file_name, full_path
+                    )
+                )
+    else:
```
choldgraf: this could just be

```
if not path_cache:
    # If we explicitly did not wish to cache, then just execute the notebook
    ntbk = execute(ntbk)
    return ntbk
```
at the beginning, and then you could de-dent all the execution logic

myst_nb/cache.py
```diff
@@ -0,0 +1,192 @@
...
+    try:
+        executor = load_executor("basic", cache_base, logger=logger)
+    except ImportError as error:
+        logger.error(str(error))
+        return 1
+    result = executor.run_and_cache(filter_pks=pk_list or None)
+    return result
+
+
+def _relative_file_path(file_path):
```
choldgraf: couldn't this just be `Path().relative_to(file_path)` or something close to that?

myst_nb/parser.py
choldgraf: thinking back on it, I think that this was actually because of testing - maybe it's changed now, but there used to be tests that were run that didn't initialize the full Sphinx app, so this was probably a hack around that. Maybe this is no longer an issue?

tests/test_execute.py
```diff
@@ -0,0 +1,213 @@
+from myst_nb.cache import execution_cache, add_notebook_outputs
```
choldgraf: yall are so much better at testing than I am...

choldgraf:
@AakashGfude I hope you don't mind, but I also pushed a little documentation page explaining the execution configuration that you gave in this repo...

choldgraf:
Really quick first thoughts from using it:

1. Regular execution worked! woo!
2. We should remove all the outputs from our documentation `ipynb` files and use jupyter cache for building our docs
3. It would be great if we could add reporting when we execute and cache notebooks, so that users know why it is hanging on certain pages
4. When I set jupyter_cache_path to `True`, I was expecting to find a jupyter cache in `_build/.jupyter_cache`, but I didn't find it! What's the default jupyter cache location?
5. I am getting some weird behavior w/ the configuration, I have

   ```python
   jupyter_execute_notebooks = True
   ```

   and it still seems to be cacheing the notebook outputs somehow, because the site builds super quickly when it is re-built (even though I've cleared the outputs from the docs files)
6. When I add `jupyter_cache = True`, then I get the following error:

    ```
	File "/c/Users/choldgraf/Dropbox/github/forks/python/ebp/myst-nb/myst_nb/cache.py", line 185, in read_nb_output_cells
	    with open(source_path, "r") as f:
	FileNotFoundError: [Errno 2] No such file or directory: 'use/index'
	```

   It seems like an extension isn't being added to the file somewhere?

myst_nb/cache.py
AakashGfude: you mean something like this ? :- 
```
dest_path = Path(env.app.outdir)
reports_dir = str(dest_path) + "/reports"
```

myst_nb/cache.py
AakashGfude: Actually the API returns an empty array if no record is found. So length of the array needs to be checked.

myst_nb/cache.py
AakashGfude: You are right. The `if` check here is also redundant

myst_nb/cache.py
AakashGfude: Yes, this was a stupid mistake. This is also the reason why execution in docs was failing with `file not found` which you mentioned in your `first thoughts from using it` comment. I guess `if "." in nb` should work, since `added` and `removed` file lists created by sphinx does not have extensions in the list.

myst_nb/cache.py
AakashGfude: Good call. smort.

myst_nb/cache.py
AakashGfude: Sure, I will make it hidden

myst_nb/cache.py
AakashGfude: right, have used `r_file_path = Path(file_path).relative_to(Path(file_path).cwd())` instead of this function

myst_nb/cache.py
AakashGfude: have used `os.path.splitext(r_file_path.name)[0]` instead of this now

AakashGfude:
> 3. It would be great if we could add reporting when we execute and cache notebooks, so that users know why it is hanging on certain pages
What 

> 4. When I set jupyter_cache_path to `True`, I was expecting to find a jupyter cache in `_build/.jupyter_cache`, but I didn't find it! What's the default jupyter cache location?

It should be present there.  The default location is `parent` of `outdir` which is _build usually.  It was there for me while running docs. Maybe the error `No such file or directory: 'use/index'` was breaking the build for the docs and preventing caching?

> 5. I am getting some weird behavior w/ the configuration, I have
>    ```python
>    jupyter_execute_notebooks = True
>    ```
>    and it still seems to be cacheing the notebook outputs somehow, because the site builds super quickly when it is re-built (even though I've cleared the outputs from the docs files)

This should b working with the new commit. `make clean` should clear the whoile `_build` folder and remove `.jupyter_cache` along with it.

> 6. When I add `jupyter_cache = True`, then I get the following error:
>    ```
>    File "/c/Users/choldgraf/Dropbox/github/forks/python/ebp/myst-nb/myst_nb/cache.py", line 185, in read_nb_output_cells
>        with open(source_path, "r") as f:
>    FileNotFoundError: [Errno 2] No such file or directory: 'use/index'
>    ```
>    
>    
>    It seems like an extension isn't being added to the file somewhere?

This was the root of most of the evil above. Should function now.

AakashGfude:
> @AakashGfude I hope you don't mind, but I also pushed a little documentation page explaining the execution configuration that you gave in this repo...

Thank you for this.

choldgraf:
Looking better! The cache worked for me this time 🎉 

A few quick thoughts:

* We probably don't need a message for both each execution *and* completion, maybe just one to avoid cluttering the messages
* When Jupyter Cache isn't set I'm getting an error: `WARNING: The config value `jupyter_cache' has type `str', defaults to `bool'.`
* Maybe we can collapse the number of configuration keys down by one, by making `jupyter_execute_notebooks` allow two values: "force" or "auto"

What other kind of functionality should I try to do to break it? :-)

AakashGfude:
> * We probably don't need a message for both each execution _and_ completion, maybe just one to avoid cluttering the messages

You are right, Ideally, we should have something like:- 
image

the screenshot is from `jupinx` . I thought of working on that after this PR is merged, as it might require discussions. What do you think?

> * Maybe we can collapse the number of configuration keys down by one, by making `jupyter_execute_notebooks` allow two values: "force" or "auto"

I agree there are too many variables. this one sounds good. will do a commit for it

> What other kind of functionality should I try to do to break it? :-)
One I think is to check if the reports are getting generated. By having one notebook failing execution with an error.

choldgraf:
Re: the last point, it wasn't an error, just a warning, so there was no trace back. Sphinx will throw a warning if the value of a config key is of a different type from the default value

AakashGfude:
> Re: the last point, it wasn't an error, just a warning, so there was no trace back. Sphinx will throw a warning if the value of a config key is of a different type from the default value

This is weird, I am not able to reproduce and not able to see in the code where the error is. Maybe will need your help in that, can you grep that variable and see if its set to a string somewhere in your system?

Also, about this point 
> Maybe we can collapse the number of configuration keys down by one, by making > jupyter_execute_notebooks allow two values: "force" or "auto"

It won't handle the case where users don't want to execute at all.

choldgraf:
Yeah - "off" seems fine if we don't want the sphinx warning about types...otherwise `None` or `False` would work...

AakashGfude:
@choldgraf Have done the necessary changes, and `jupyter_execute_notebooks` can now have  "force", "auto", "off" values. I have also updated the docs.

choldgraf:
Agree re: the jupinx loading bars, I think that style would be best!

A few more thoughts:

* When I have `jupyter_execute_notebooks` set to "force", but jupyter_cache to `True`, then it seems to keep using the cache. This is a bit confusing as I assume it would force re-execution each time.
* What about we collapse the config a bit further. `jupyter_execute_notebooks` take 4 possible options: `auto` (default), `force` (always execute), `cache` (equiv to `jupyter_cache = True`), and `off`. Then, `jupyter_cache` is always used *just* if a person wants to point to a different cache location.
* Execution exclusion doesn't seem to work for me, I put `execution_excludepatterns = ['*glue*']`, and `docs/use/glue.ipynb` was still executed. Does it not support glob-like patterns? (if not, can we support that?)
* I noticed that the "Executing notebook" log message doesn't show up if we don't use a jupyter cache - is there a reason *not* to tell people it's executing if we aren't cacheing? I think we could say "executing" if `jupyter_cache = False`, and "executing and cacehing" if `jupyter_cache = True`.

AakashGfude:
> * What about we collapse the config a bit further. `jupyter_execute_notebooks` take 4 possible options: `auto` (default), `force` (always execute), `cache` (equiv to `jupyter_cache = True`), and `off`. Then, `jupyter_cache` is always used _just_ if a person wants to point to a different cache location.

So, in this case, `auto` will use the cache as well right. What will be the difference then with `cache`? And the person will use the `jupyter_cache` variable only when he wants to point to a different location?

> * Execution exclusion doesn't seem to work for me, I put `execution_excludepatterns = ['*glue*']`, and `docs/use/glue.ipynb` was still executed. Does it not support glob-like patterns? (if not, can we support that?)

Yes, at present it was just doing a substring check of `execution_excludepatterns` with the path of files. I will have a look at how to support glob-like patterns.

> * I noticed that the "Executing notebook" log message doesn't show up if we don't use a jupyter cache - is there a reason _not_ to tell people it's executing if we aren't cacheing? I think we could say "executing" if `jupyter_cache = False`, and "executing and cacehing" if `jupyter_cache = True`.

I saw some logs about execution from nbclient as well. But, I just realized that it shows the log only for markdown files. And it executes markdown files as well. I will supress that by configuring it. and give logs as mentioned

AakashGfude:
@choldgraf have updated the functionalities, feel free to test it and let me know your thoughts

choldgraf:
I just gave it another shot, and it worked really nicely 👍 - I tried all the different config options, and it worked really well, great job! I pushed a small extra bit to the documentation and also cleared out the outputs in our `ipynb` files in the docs so that we can actually put this to the test on a daily bases :-)

One thing that I foresee potentially confusing people in the future: we should provide some kinds of guidance for how to "clear the cache". E.g., do they delete `_build`? What if they only want to delete the pages to trigger a re-build, but don't want to remove the cache artifacts? That kinda thing...I don't know that it's super actionable right now though, because I'm not sure what the right path forward is, maybe something to keep an eye on as people use it.

I think that this one is nearly ready to go - we should treat this functionality as "late-alpha, early beta" for a while, so there's time for users to give the API a shot and provide feedback, squash bugs, etc. But I think it's ready to go from a user functionality and docs standpoint :shipit: 

@chrisjsewell what do you think?

tests/test_execute.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@pytest.fixture()
+def mock_document(tmp_path) -> nodes.document:
+    settings = OptionParser(components=(RSTParser,)).get_default_values()
+    document = new_document("notset", settings=settings)
+    document.settings.env = MockEnv(tmp_path)
+    with sphinx_domains(document.settings.env):
+        yield document
+
```
chrisjsewell: This can all be moved to `conftest.py` and merged with the largely duplicated code there (if you didn't know, you can use test fixtures defined in conftest.py in any test file with the same root path)

chrisjsewell:
Looks good to me 👍

tests/test_execute.py
choldgraf: @AakashGfude do you wanna try making this simplification to the tests, and then if they're happy we can merge?

AakashGfude: @choldgraf @chrisjsewell  sure had separated it to play around while not affecting other tests. Thanks for the reminder.

AakashGfude:
> One thing that I foresee potentially confusing people in the future: we should provide some kinds of guidance for how to "clear the cache". E.g., do they delete `_build`? What if they only want to delete the pages to trigger a re-build, but don't want to remove the cache artifacts? That kinda thing...I don't know that it's super actionable right now though, because I'm not sure what the right path forward is, maybe something to keep an eye on as people use it.

What I think we can have is have targeted clean methods like `make clean` to clear eveything. `make clean-html` to clear html outputs while preserving the cache. `make clear-cache` to clear `.jupyter_cache` only(We will have to take custom jupyter_cache path provided into account).

deleting the pages from the cache to trigger a re-build without removing the artifacts is something we need to think about.

choldgraf:
fantastic! I'll plan on giving this a merge when I'm back at my computer (prob tomorrow morning)

chrisjsewell:
You could then bump the version to `0.5.0a1` and create a release: https://github.com/ExecutableBookProject/MyST-NB/releases, and it will automatically be published to PyPi as a pre-release during the triggered travis-CI

choldgraf:
alright let's 🚢 it!

choldgraf:
OK I merged, bumped the version on master, and also added a release: https://github.com/ExecutableBookProject/MyST-NB/releases/tag/0.5.0a1

chrisjsewell:
> I bumped master before making the official release on github, so I triggered a travis re-build because I'm assuming that needs to happen once a release already exists

Not sure if I understand this, but when you make a release that triggers Travis (or Gihub actions as we may move to soon), so you don't need to manually trigger. But anyway its all sorted 👍 

FYI, if ppl didn't know, you now have the pre-release on https://pypi.org/project/myst-nb/#history, whereby this will **not** be installed doing `pip install myst-nb`, it will only be installed if you specifically specify `pip install myst-nb==0.5.0a1`

PR #57, opened 2020-03-09
using jupyter-sphinx output converted

choldgraf:
This is an attempt at using the jupyter-sphinx output node function, and having it write to disk instead of inlining images.

It works as expected, with one problem, which is that we had to move the mimebundle transform from `post-transform` to `transform`...see the code comment for more context


myst_nb/__init__.py
```diff
@@ -82,17 +81,8 @@ def visit_element_html(self, node):
...
-        atts = {"src": f"data:image/png;base64, {node['uri']}", "alt": f"{node['alt']}"}
-        self.body.append(self.emptytag(node, "img", "\n", **atts))
-
-    app.add_node(
-        CellImageNode, override=True, html=(visit_cell_image, lambda self, node: "")
-    )
-
     # Register our post-transform which will convert output bundles to nodes
-    app.add_post_transform(CellOutputsToNodes)
+    app.add_transform(CellOutputsToNodes)
```
choldgraf: I think this is the biggest problem we'd be introducing.

Before we were using a `post-transform`, so that the output nodes would *only* be chosen once we knew the builder type. However, this raised the issue that Sphinx runs an `ImageCollector` transform *before* the `post_transforms` step, and so if any image nodes were added in the `post_transform` step, then Sphinx would miss them.

So this PR moves the output mimebundle conversion to a transform instead of a post-transform. @chrisjsewell or @akhmerov any thoughts on how much of a problem this would cause?

I suppose our other option would be to try and re-create whatever the ImageCollector transform is doing, but as a post-transform (or maybe we add the ImageCollector itself as a post-transforM?)

Here's the ImageCollector transform: https://github.com/sphinx-doc/sphinx/blob/af62fa61e6cbd88d0798963211e73e5ba0d55e6d/sphinx/environment/collectors/asset.py#L33

choldgraf:
This also fails the tests because it now needs Sphinx-specific information at *parse* time, instead of just at the transform step. Need to figure out how to add that in to the tests somehow

myst_nb/__init__.py
chrisjsewell: The reason for it being a post-transform is that you need to output different mime types for different builders (e.g. HTML and latex). Moving this to a transform means that you can’t build multiple output types, because you now are hard coding the stored doctrees to being for a certain output type. A post-transform is transient, and so does not have this issue. 

For this reason, I’m against this change.

choldgraf: Yeah - that makes sense to me. I think the challenge is that Sphinx does some clever image collecting and path renaming etc beforehand. So, perhaps it's possible to mimic the same behavior but without needing the post-transform?

chrisjsewell: I think replicating the collector step is the better approach: not ideal, but doesn’t break the key concept of the stored doctrees being output agnostic. Note nbsphinx uses the `only` directive to store outputs for both HTML and latex. It’s a possible solution, but still I really don’t like it, since it can break later processing of reference/target resolution (e.g. if you label your output figures)

choldgraf:
ah ha, I figure it out - it turns out that you can just manually call the `ImageCollector` class, and that's what this PR now does at post-transform time!

I am not sure how to fix the tests. I think I need to do something like what @chrisjsewell has done with MockEnv but this is a bit outside of my pytest skills. Can anybody advise?

chrisjsewell:
Tests now check that the correct files were written to disk 😄

choldgraf:
amazing, man I owe you like 10 beers

Issue #58, opened 2020-03-10
Figure out how to keep a consistent Mistletoe AST before the docutils render


choldgraf:
In https://github.com/ExecutableBookProject/MyST-Parser/issues/113 we realized that there would be some unexpected behavior as a result of parsing each cell as a separate "document". e.g., footnote definitions would not be collected across multiple cells.

We should figure out how to get around this

@mmcky mentioned this would be particularly problematic for him!

mmcky:
thanks for raising this issue @choldgraf. I think this is mainly a problem for the use case of supporting books in general. I think `footnotes` will need to be collected and resolved across cells as it is reasonable to expect a file to represent chapters / sections etc. 

How do we resolve `references` across notebooks (at the project level)?  I guess the issue here is that `references` such as `bibtex` is global across `notebooks` while footnotes are a local definition within each `notebook`?

chrisjsewell:
With the shiny new `mistletoe-ebp` API (see https://mistletoe-ebp.readthedocs.io/en/latest/using/intro.html#programmatic-use), I think there could be a better way to parse the notebook, more inline with how it will occur for the text based representation, along the lines of:

1. Convert the full notebook to the 'intermediate' mistletoe/markdown AST (before inline text is processed)
2. Walking this AST to 'expand' `SpanContainer` (see https://mistletoe-ebp.readthedocs.io/en/latest/using/intro.html#the-parse-process)
2. Applying the `SphinxRenderer` once on the full AST (rather than for each cell as is done now).

This would involve creating a specific `CodeCell` mistletoe token, then adding an extra render method to `SphinxRenderer` that knows how to deal with it, basically using the code from https://github.com/ExecutableBookProject/MyST-NB/blob/d089de3c1140f72daf7c579be7637164ee548028/myst_nb/parser.py#L98

This logic could be problematic: https://github.com/ExecutableBookProject/MyST-NB/blob/d089de3c1140f72daf7c579be7637164ee548028/myst_nb/parser.py#L89

As we have already discussed in #30

Ideally we would also need to work out a way to incorporate #21. A first step to achieving this would be  ExecutableBookProject/mistletoe-ebp#9

PR #59, opened 2020-03-11
Refactor parsing process

chrisjsewell:
This is a draft implementation of the parsing process outlined in #58


chrisjsewell:
thoughts @choldgraf @mmcky ?

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+            find_spans=SphinxNBRenderer.default_span_tokens,
+        )
+        # TODO hook in the document.reporter here,
+        # to e.g. report on duplicate link/footnote definitions
+        set_parse_context(parse_context)
+
+        for cell_index, nb_cell in enumerate(ntbk.cells):
+
+            # Skip empty cells
+            if len(nb_cell["source"]) == 0:
```
choldgraf:
```suggestion
            if len(nb_cell["source"].strip()) == 0:
```
a bug that we need to fix anyway, I think :-)

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+        # de-serialize the notebook
+        ntbk = nbf.reads(inputstring, nbf.NO_CONVERT)
+
+        # This is a contaner for top level markdown tokens
+        # which we will add to as we walk the document
+        mkdown_tokens = []
+
+        # first we ensure that we are using a 'clean' global context
+        # for parsing, which is setup with the MyST parsing tokens
+        parse_context = ParseContext(
```
choldgraf: Is ParseContext a new addition to mistletoe from us?

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+                    for line in lines
+                ]
+
+                # parse the markdown;
+                # at this point span/inline level tokens are not yet processed, but
+                # link/footnote definitions are collected/stored in the global context
+                # TODO here it would be ideal to somehow include the cell index
+                # in the `position` attribute of each token
+                # and utilise in within the document.reporter
+                mkdown_tokens.extend(tokenize_block(lines))
```
choldgraf: And the *definitions* are always block-level, yeah?

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+
+                # TODO think of a way to implement the previous
+                # `if "hide_input" in tags:` logic
+
+            elif nb_cell["cell_type"] == "code":
+                # here we do nothing but store the cell as a custom token
+                # TODO here index would instead be part of the position attribute
+                mkdown_tokens.append(NbCodeCell(cell=nb_cell, index=cell_index))
```
choldgraf: Could this token keep information about the tags in the cell? And then those tags could be propagated later to deal w/ things like `hide_input`?

I wonder if we could then say "markdown hiding/toggle/etc behavior should be triggered with directives, whereas code cell behavior can be triggered w/ tags"

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+
+        # Now all definitions have been gathered, we walk the tokens and
+        # process any inline text
+        # TODO maybe make this a small function in mistletoe/myst_parser
+        for token in mkdown_tokens + list(
+            get_parse_context().foot_definitions.values()
+        ):
+            for result in list(token.walk(include_self=True)):
+                if isinstance(result.node.children, SpanContainer):
+                    result.node.children = result.node.children.expand()
```
choldgraf: Just to be clear, this is still pre-docutils right? I just see "node" and "children" which sounds like docutils, but maybe that's just generic AST speak...

myst_nb/parser.py
```diff
@@ -21,6 +27,94 @@ class NotebookParser(MystParser):
...
+                # TODO this is going to be changed in mistletoe, to just be
+                # lines = SourceLines(lines)
+                if isinstance(lines, str):
+                    lines = lines.splitlines(keepends=True)
+                lines = [
+                    line if line.endswith("\n") else "{}\n".format(line)
+                    for line in lines
+                ]
+
+                # parse the markdown;
```
choldgraf: Will we get a significant performance hit doing it this way?

myst_nb/parser.py
```diff
@@ -39,69 +131,43 @@ def parse(self, inputstring, document):
...
-                renderer.render(document)
-        else:
-            renderer.render(document)
-
-    # If a code cell, convert the code + outputs
-    elif cell["cell_type"] == "code":
+
+        # render the Markdown AST to docutils AST
+        renderer = SphinxNBRenderer(
+            parse_context=parse_context, document=document, current_node=None
```
choldgraf: I can't figure out where, but before now the span-level blocks need to have been parsed right?

myst_nb/parser.py
```diff
@@ -39,69 +131,43 @@ def parse(self, inputstring, document):
...
+        )
+        renderer.render(markdown_doc)
+
+
+class NbCodeCell(BlockToken):
+    def __init__(self, cell, index):
+        self.cell = cell
+        self.index = index
```
choldgraf: tags =?

or

metadata=?

chrisjsewell:
> This seems like a good way to do things! I added a few comments and questions, and still need to wrap my head around the new `Document` behavior, but looks to be a step in the right direction for sure.

Cheers I'll go through them but firstly, if you haven't already, make sure to read https://mistletoe-ebp.readthedocs.io/en/latest/using/intro.html#usage, which explains/illustrated a lot of the 'new' mistletoe parsing process

myst_nb/parser.py
chrisjsewell: Yes, it is a consolidation of a number of aspects of the original code that utilised global variables. Now we contain them in a single place, that is also thread-safe to use in multithreaded applications (e.g. for LSP) 😄

myst_nb/parser.py
chrisjsewell: Yes [LinkDefinition](https://mistletoe-ebp.readthedocs.io/en/latest/api/core_block_tokens.html#linkdefinition) and [Footnote](https://mistletoe-ebp.readthedocs.io/en/latest/api/extension_tokens.html#footnote) are block tokens

myst_nb/parser.py
chrisjsewell: Here we are just storing the entire notebook cell, i.e. 

```json
{
      "cell_type": "code",
      "execution_count": 1,
      "metadata": {"tags": []},
      "source": [],
      "outputs": []
}
```

myst_nb/parser.py
chrisjsewell: Yes absolutely no docutils stuff yet. The `mkdown_doc` we create could actually be used by any renderer into, assuming we implemented a `render_nb_cell` method on it.

myst_nb/parser.py
chrisjsewell: No this is following the exact process that is used 'internally' in mistletoe

myst_nb/parser.py
chrisjsewell: see comment above, the cell is the entire code cell

myst_nb/parser.py
chrisjsewell: The span level blocks are parsed above in `result.node.children = result.node.children.expand()` (see https://mistletoe-ebp.readthedocs.io/en/latest/using/intro.html#the-parse-process)

myst_nb/parser.py
chrisjsewell: For code cells then, you can utilise the metadata.

For the markdown hiding/toggle/etc behaviour yeh I think the question is what do you actually want to hide? 
- For admonitions (notes, warnings, etc) then certainly just use the `:class:` option, as shown in https://github.com/ExecutableBookProject/sphinx-togglebutton#usage
- For sections, it would be interesting to think about if, at the `SphinxRenderer` level, you want to extract classes from the end of header text. For example, you could write something like

```markdown
# A header with a class {.toggle}
```

The in [render_header](https://github.com/ExecutableBookProject/MyST-Parser/blob/c64f7068ba786a79561adb0017194ebc8bf15ba3/myst_parser/docutils_renderer.py#L299) you would look for these and apply the class to the section (would `sphinx-togglebutton` currently handle this)

choldgraf: I was thinking about just implementing a `{toggle}` directive that would wrap everything inside as "toggle-able". The only restriction (I think) would be that it couldn't have a header inside.

choldgraf: but I think either way, we don't need to make "toggle entire sections w/ headers" a blocker on this

chrisjsewell:
@choldgraf I've bubbled up the improvements discussed to mistletoe/myst_parser addressing most of the TODOs 👍

chrisjsewell:
@choldgraf @mmcky this is ready for merging; check out a footnote in a notebook 😄 https://122-241268229-gh.circle-artifacts.com/0/html/use/basic.html#note

chrisjsewell:
Hmm, noticed that widgets aren't showing up though: https://122-241268229-gh.circle-artifacts.com/0/html/use/interactive.html#ipywidgets.

I still have the code in there for inserting widget state: https://github.com/ExecutableBookProject/MyST-NB/blob/ec652e93ddc9e79f345e283410dda66713644594/myst_nb/parser.py#L132

So I'm not sure what correction need to be made??

chrisjsewell:
> Hmm, noticed that widgets aren't showing up though.

Fixed in 46e773e 😄: https://124-241268229-gh.circle-artifacts.com/0/html/use/interactive.html

myst_nb/parser.py
```diff
@@ -21,6 +30,92 @@ class NotebookParser(MystParser):
...
+                    metadata={"cell_index": cell_index},
+                    standardize_ends=True,
+                )
+
+                # parse the source markdown text;
+                # at this point span/inline level tokens are not yet processed, but
+                # link/footnote definitions are collected/stored in the global context
+                mkdown_tokens.extend(tokenize_block(lines))
+
+                # TODO think of a way to implement the previous
```
choldgraf: Is this a regression then? (currently `hide_input` etc does work, so will this remove that functionality?)

choldgraf: it seems below that we *do* handle tags in the Renderer, so maybe this comment just needs to be removed?

chrisjsewell: This comment is **only** referring to `hide_input` tags for markdown cells. But as we discussed, we can now handle this by directives. So I can remove this comment if you are happy with losing that functionality.

chrisjsewell: i.e. we no longer implement this code: https://github.com/ExecutableBookProject/MyST-NB/blob/2207e30a63928c4d96c46a395512608595735fa5/myst_nb/parser.py#L89

choldgraf:
```suggestion
                # TODO for md cells, think of a way to implement the previous
```

choldgraf: does that work?

chrisjsewell: yep fine by me

choldgraf:
OK cool - I'll hit the green button when tests are happy 👍 this looks really nice, and the footnotes are awesome :-)

choldgraf:
boom!

PR #61, opened 2020-03-12
Update installation instructions

chrisjsewell:


chrisjsewell:
See https://132-241268229-gh.circle-artifacts.com/0/html/index.html#installation

PR #66, opened 2020-03-13
adding glue and paste functionality

### LINKED ISSUE #4: IPYNB parsing: auto-register scrapbook scraps somehow
An idea I just had when thinking about how we can reference content from one notebook to another - what if as a part of the Sphinx parser we automatically generated docutils targets from any scraps that are in a notebook during the parsing process. That way references to those targets could already exist and users could reference them with a role, similar to what we do with `:ref:`. I'm imagining something like:

## In a notebook

Somebody writes an analysis that generates a plot they'd like to include elsewhere in their docs. They run `scrapbook.glue('myplot')`

## In a MyST document

Somebody wants to include that "scrap" from the notebook. They only need to do something like

```
{scrap}`myplot`
```
and Sphinx just looks up `myplot` against a list of scraps that it has found across any parsed notebook.

choldgraf:
This adds a prototype functionality for "glue and paste" with MyST-NB. It closes #4 

It doesn't yet have tests because I want feedback on the API and functionality first!

You glue things into a notebook's metadata like this:

```python
from myst_nb import glue
glue("your-key", an_object)
```

And it will run IPython's display on the object, then store the mimebundle
at the key you specify.

When the notebooks are parsed with `MyST-NB`, it builds up a registry of all
the keys across all notebooks, so that you can then refer to them in the following ways:

You can paste it in markdown with a directive like this:

````
```{paste} your-key
```
````

Or you can add it in-line like this:

```
{paste}`your-key`
```
optionally:

```
{paste}`your-key:format-string`
```

Here's a docs page demo-ing how this works: https://141-241268229-gh.circle-artifacts.com/0/html/use/glue.html

I think there are still some other things we'd like to add, like adding captions to figures somehow etc, but I'm not sure whether that should be done with `glue` functions, or if it should be handled by the `MimeOutputBundle` transform.

Would love to hear thoughts! cc @chrisjsewell @mmcky @jstac and @AakashGfude (that's the group that first discussed this feature)

### CODE REVIEW

File: myst_nb/glue.py [hunk 1]
```diff
@@ -0,0 +1,173 @@
...
+        for output in cell["outputs"]:
+            meta = output.get("metadata", {})
+            if "scrapbook" in meta:
+                this_key = meta["scrapbook"]["name"].replace(mime_prefix, "")
+                if key == this_key:
+                    bundle = output["data"]
+                    bundle = {this_key: val for key, val in bundle.items()}
+                    outputs.append(bundle)
+    if len(outputs) == 0:
+        print(f"Warning: did not find key {this_key} in notebook {path_ntbk}")
```
chrisjsewell:
```suggestion
        SPHINX_LOGGER.warning(f"Did not find key {this_key} in notebook {path_ntbk}")
```

myst_nb/glue.py
```diff
@@ -0,0 +1,173 @@
...
+        return
+    if len(outputs) > 1:
+        print(
+            f"Warning: multiple variables found for key: {this_key}. Returning first value."
+        )
```
chrisjsewell:
```suggestion
        SPHINX_LOGGER.warning(
            f"Multiple variables found for key: {this_key}. Returning first value."
        )
```

myst_nb/parser.py
```diff
@@ -124,14 +125,40 @@ def parse(self, inputstring, document):
         except AttributeError:
             pass
 
-        # Write the notebook's output to disk
+        # Remove all the mime prefixes from "glue" step.
```
chrisjsewell: maybe add a configuration option to turn this functionality on/off

chrisjsewell:
In general, I think it could be good; as long as the implementation doesn't become too complex/error prone

chrisjsewell:
> I think there are still some other things we'd like to add, like adding captions to figures somehow 

We definitely need to get round to this sooner rather than later. Adding captions, referencable labels and other general formatting (size, alignment, etc) to code outputs is a major aspect of what we want to achieve IMO

myst_nb/parser.py
choldgraf: This is needed in order to ensure that the files actually get written to disk. The jupyter-sphinx writer step only writes image files that it has prefixes for. So if we have the `nteract/application` stuff in there it'll skip them.

Another option would be to try and upstream some logic to `jupyter-sphinx` that'll check for a prefix...

choldgraf:
@chrisjsewell well, we have a directive that we *could* do other things with. Perhaps we can pass some more metadata through the `OutputBundle` nodes, and then we can add functionality for things like captions in figures etc in the post-transform step?

myst_nb/parser.py
chrisjsewell: I meant to turn off the entire glue functionality, but obviously anythingupstreamed tojupyter-sphinx is probably good

choldgraf: @chrisjsewell ah that's a good point, agreed

choldgraf: @akhmerov would you be +1 on an addition to the "write_outputs" function that would allow us to add a "prefix" to any outputs that it would also search for? That way `image/png` would work as well as `myprefix/image/png`.

chrisjsewell:
> @chrisjsewell re: captions,references, etc we do have a directive that we _could_ do other things with. Perhaps we can pass some more metadata through the `OutputBundle` nodes, and then we can add functionality for things like captions in figures etc in the post-transform step?

Yeh definitely, something like that 👍

chrisjsewell:
With the paste role/directive we also have the opportunity to exert a lot of control over the output. I could envisage creating a [sphinx domain](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html), e.g.

````
```{paste:literal} key1
```

```{paste:figure} key2
:name: label

A caption...
```

```{paste:subfig}
- key3
- key4
```
````

**Comment by chrisjsewell** [comment]
> With the paste role/directive we also have the opportunity to exert a lot of control over the output. I could envisage creating a sphinx domain, e.g.

If we go this route, although I like the word `paste`, it could be a bit long, so perhaps `nb` would be better?

```rst
:nb:text:`your-key:format-string`
```

myst_nb/glue.py
```diff
@@ -0,0 +1,173 @@
...
+from sphinx.util import logging
+
+from .parser import CellNode, CellInputNode, CellOutputBundleNode
+
+SPHINX_LOGGER = logging.getLogger(__name__)
+
+mime_prefix = "application/papermill.record/"
+
+
+def glue(name, variable):
```
chrisjsewell: For things like figures, we probably want to show them in the notebook, at the same time as 'glueing' them, e.g. `glue("key", fig, display=True)`.

a) This would probably be used at the same time as tagging the cell with `remove_output`
b) In this use case we probably wouldn't want to prefix the mime types, because then we would have to store duplicate prefixed/non-prefixed outputs, so we would want to indicate whether the mime types had been prefixed, e.g.

`metadata["scrapbook"] = dict(name=name, mime_prefix=None)`

choldgraf: yeah - my thinking was that people probably have already shown the figure, since it gets shown by default when you create the matplotlib object. At least when I've used scrapbook, I *always* end up having two figures plotted side-by-side, which is why I didn't feel the need to implement this in the current pass (since the act of "gluing" a plot is orthogonal to the act of "displaying" a plot). Do you feel strongly that this needs to be possible?

choldgraf:
> :nb:text:`your-key:format-string`

Interesting idea - I do like the idea of giving more specific commands when calling the role/directive (e.g. `{nb:text}` and `{nb:fig}`), though it's also a bit more complex to remember...and probably outside of my sphinx knowledge. I am happy to merge this as a super-alpha module, if you want to hack it around and improve the API. What do you think?

myst_nb/glue.py
chrisjsewell: Yes matplotlib is a bit of an outlier in this respect, in that basically it doesn't use the `IPython` mechanics properly. All other packages, like pandas, (correctly) don't have this 'auto-display' behaviour, and so this is where you would use it.
(you can also turn off matplotlib's auto show I think: https://stackoverflow.com/questions/18717877/prevent-plot-from-showing-in-jupyter-notebook)

chrisjsewell:
> I am happy to merge this as a super-alpha module to get a feel for how the API works

Yeh, address this:
https://github.com/ExecutableBookProject/MyST-NB/pull/66#discussion_r392030315,
and fix the tests, then probably we can merge and add the role/directive domain idea as an `enhancement` issue

choldgraf:
sounds good - I'll add back in the display stuff and merge it in after some tests that pass

myst_nb/glue.py
chrisjsewell: I think you accidentally reverted this in your force push

myst_nb/glue.py
chrisjsewell: ditto

myst_nb/__init__.py
```diff
@@ -35,6 +36,11 @@ def update_togglebutton_classes(app, config):
         config.togglebutton_selector += f", {selector}"
 
 
+def init_glue_cache(app):
+    if not hasattr(app.env, "glue_data"):
+        app.env.glue_data = {}
```
chrisjsewell: It would be nice to have 'external' access to this dict, to see what keys we have stored. Maybe we can write a small function in myst-nb, that `ExecutableBookProject/cli` could utilise (outside of sphinx) to show the currently stored keys:

```python
from pathlib import Path
import pickle

def get_glue_data(doctreedir):
    path = Path(doctreedir, "environment.pickle")
    env = pickle.loads(path.read_bytes())
    return env.glue_data
```

myst_nb/glue.py
```diff
@@ -0,0 +1,177 @@
...
+            # First check if we have both key:format in the key
+            parts = paste_node.key.rsplit(":", 1)
+            if len(parts) == 2:
+                key, formatting = parts
+            else:
+                key = parts[0]
+                formatting = None
+
+            if key not in glue_data:
+                raise KeyError(
```
chrisjsewell: I think this should use the sphinx logging mechanics, to also report the `paste_node` source file and line number

myst_nb/glue.py
```diff
@@ -0,0 +1,177 @@
...
+# Role and directive for pasting
+class Paste(SphinxDirective):
+    required_arguments = 1
+    final_argument_whitespace = True
+    has_content = False
+
+    option_spec = {"id": directives.unchanged}
+
+    def run(self):
+        return [PasteNode(self.arguments[0], "directive")]
```
chrisjsewell: I think you need to add the source path / line number to the node here, for later use in reporting (have a look at other sphinx directives)

myst_nb/glue.py
```diff
@@ -0,0 +1,177 @@
...
+                    # If formatting is specified, see if we have a number of some kind
+                    if formatting:
+                        try:
+                            newtext = float(text)
+                            text = f"{newtext:>{formatting}}"
+                        except ValueError:
+                            pass
+                    out_node = nodes.inline(text, text, classes=["pasted-text"])
+                else:
+                    SPHINX_LOGGER.warn(
```
chrisjsewell: ditto reporting source file / line number

myst_nb/glue.py
```diff
@@ -0,0 +1,177 @@
...
+                        f"Couldn't find compatible output format for key {key}"
+                    )
+            # Directives will have the whole output chunk deposited and rendered later
+            elif paste_node.kind == "directive":
+                output_node = CellOutputBundleNode(outputs=[output])
+                out_node = CellNode()
+                out_node += CellInputNode()
+                out_node += output_node
+            else:
+                raise ValueError(
```
chrisjsewell: ditto reporting source file / line number

myst_nb/glue.py
chrisjsewell: Also, should we be raising exceptions, that block the whole build, or just reporting warnings/errors

myst_nb/__init__.py
chrisjsewell: Or alternatively you store this as a separate JSON file in the `doctreesdir`.

choldgraf:
I think I just addressed all your comments in the latest commit - LMK if that looks correct to you. I'm still not sure how to handle line numbers with the ipynb files, but in the meantime I'm just removing the line number from warnings with an `.ipynb` extension, and we can figure that out later.

We still need to add tests for this. At the least:

* Ensure `glue` works as expected, puts in the correct mime information etc
* Ensure the role + directive grab the correct glued items and insert them in the doctree
* Ensure the glue cache is correct after a build

Anything else?

choldgraf:
(also, I am not really sure how to test this...it seems that our test infrastructure is just running the parser + transform on a single file, whereas this requires us to test something that collects across multiple files in a sphinx site)

chrisjsewell:
I’m on it...

choldgraf:
@chrisjsewell I just pushed another commit that adds some very basic tests that don't quite work. It uses teh regression framework you have, but it doesn't yet do multi-page testing, and also I can't figure out how to register the new roles+directives in sphinx.

Happy for you to clobber those changes if you like :-)

myst_nb/glue.py
```diff
@@ -0,0 +1,234 @@
...
+        self.key = key
+        self.kind = kind
+        self.location = location
+        super().__init__("", **attributes)
+
+
+class PasteNodesToDocutils(SphinxTransform):
+    """Use the builder context to transform a CellOutputNode into Sphinx nodes."""
+
+    default_priority = 700
```
chrisjsewell: We should add a comment of justification for why 700 was chosen, and ditto for `CellOutputsToNodes`. Also should one be a higher priority than the other?

choldgraf: Ummmm...that makes a very strong assumption that I have logical reasoning behind my decisions...

But I did a [quick wikipedia search](https://en.wikipedia.org/wiki/700_(number)) and this is why I've decided that the number is 700:

> 700 (seven hundred) is the natural number following 699 and preceding 701. It is the sum of four consecutive primes (167 + 173 + 179 + 181).

chrisjsewell: 😆

chrisjsewell:
tests are done

chrisjsewell:
We could add a full sphinx integration test (like in myst-parser), but I'll save that for a later date

choldgraf:
You are amazing, much appreciated 🙏

choldgraf:
Are you +1 on merging this, and opening an issue to discuss the role / directive names etc?

chrisjsewell:
> Are you +1 on merging this, and opening an issue to discuss the role / directive names etc?

Once you resolve https://github.com/ExecutableBookProject/MyST-NB/pull/66#discussion_r392544779, yes 😁

myst_nb/glue.py
chrisjsewell: Yeh `PasteNodesToDocutils` needs to be run before `CellOutputsToNodes` (see 
72a5595)

choldgraf:
oops - I just saw this...thanks for changing the priority

chrisjsewell:
No problem, I jumped in to merge it before you, so that I could add a proper commit message 😉

Issue #64, opened 2020-03-12
Code cell prefixes


chrisjsewell:
As per https://nbsphinx.readthedocs.io/en/0.5.1/code-cells.html#Code-Cells,
it would be nice to have the option of including the cell prefixes `[1]: ` (or in a format specified in the conf.py)

@akhmerov I don't think this is possible with `jupyter-sphinx` at present?

More generally, we should probably also being looking at nbsphinx (and talking to mgeiger) and thinking about what the synergies are.

akhmerov:
No, it's not possible with jupyter-sphinx at the moment. I am wondering also whether it is a desirable feature in the context of building a book/documentation. What use cases do you imagine?

akhmerov:
I mean in a notebook these tell you about the sequence of executions, but in a book all executions should be sequential.

Also imagine if some inputs are hidden and some shown. Should the numbering include the hidden ones or not?

chrisjsewell:
> I am wondering also whether it is a desirable feature in the context of building a book/documentation

Well I'm not necessarily as bothered about the numbering, its more to have a clear distinction of what are inputs and what are outputs, particularly for tutorials. 

For example, here is one I have created with IPyPublish, https://ase-notebook.readthedocs.io/en/latest/user_guide/tutorial_intro.html. I think this is nicer to read than e.g. this: https://myst-nb.readthedocs.io/en/latest/use/basic.html#code-cells-and-outputs, where all the blocks kind of merge into each other

chrisjsewell:
Also, in the latter example, it would not be clear what is executed code and what is static code, e.g.

````markdown
```python
a = 1
```

```{jupyter-execute}
b = 2
```
```

would both look the same

chrisjsewell:
Or here I just use `In:` and `Out:`  https://aiida-crystal17.readthedocs.io/en/develop/user_guide/calc_main.html

chrisjsewell:
I could even envisage referencing these cells, e.g.

```rst
In :cell:in:`1` we calculated something, now in :cell:in:`5` we calculate something else.
````

where the reference would link to the prefix

akhmerov:
Both points you brought up are relevant to address, but I'm not sure whether cell numbering does that in the best way. `In:` and `Out:` seems better, and of course everything may be configurable.

---

[...]

PR #68, opened 2020-03-13
Add CSS; left green border to input cells

chrisjsewell:
As discussed in #64

### DIFF
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -20,6 +20,8 @@ div.cell div.cell_input {
   padding-left: 0em;
   border: 1px #ccc solid;
   background-color: #f7f7f7;
+  border-left-color: green;
+  border-left-width: medium;
 }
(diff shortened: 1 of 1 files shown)


chrisjsewell:
See: https://147-241268229-gh.circle-artifacts.com/0/html/use/basic.html#code-cells-and-outputs

chrisjsewell:
I think this already on the 'TODO' list, but we still need to standardize the CSS across myst-nb and jupyter-sphinx

choldgraf:
re: standardizing, for sure. I've got https://github.com/jupyter/jupyter-sphinx/pull/107 open, which will let us remove another chunk of code from myst-nb and just use the upstream. That is bogged down a bit right now on figuring out how to re-work the testing though, because now we need an actual sphinx build instead of just a docutils parse :-/

I think this looks quite nice! Good thinking to add a lightweight visual cue for cells

PR #73, opened 2020-03-14
Implement `NbGlueDomain`

### LINKED ISSUE #70: Consider using a sphinx domain for paste roles/directives
With the paste role/directive introduced in b46c2f1f5b8d155866e531a1199188fd588517bd we also have the opportunity to exert a lot of control over the output. I could envisage creating a [sphinx domain](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html), e.g.

````
```{paste:literal} key1
```

```{paste:figure} key2
:name: label

A caption...
```

```{paste:subfig}
- key3
- key4
```
````

_Originally posted by @chrisjsewell in https://github.com/ExecutableBookProject/MyST-NB/pull/66#issuecomment-598534649_

### PR DESCRIPTION
**Pull request by chrisjsewell**
closes #70 and also relates to #72 

An initial draft of reworking the paste roles/directive into a domain. The domain achieve's two things:

1. It collates all roles/directives relating to rendering of 'glue' objects. So all you need to have in the extension setup is `app.add_domain(NbGlueDomain)`.
2. It centralises access to the glue cache. This has allowed me, for example, to also track the notebook from where each glue key comes from, and remove keys related to a deleted document (see `NbGlueDomain.clear_doc`)

I have then added the `PasteFigure` directive as an example of extending the functionality of the domain for handling different output types. It allows for captions, labelling/referencing, width and alignment.
Take a look at:
https://173-241268229-gh.circle-artifacts.com/0/html/use/glue.html#using-the-nb-figure-directive
😄

### CODE REVIEW

**Comment by chrisjsewell** [comment]
In d4aa616, I have refactored the paste role and rewritten the documentation accordingly: https://175-241268229-gh.circle-artifacts.com/0/html/use/glue.html#pasting-glued-variables-into-your-page

**Comment by choldgraf** [comment]
this looks awesome - I'll take a look soon!

**Comment by chrisjsewell** [comment]
I was also thinking of changing `nb:paste` to `nb:output`. I'm not fixed on the naming though and am open to other suggestions.

**Comment by choldgraf** [comment]
took a look at the docs - the figure bit is super cool!!

A few quick thoughts:

* I wonder if the `nb:` prefix will be clear to people. It seems reasonable to me, but we should shop that around with the team to see if it is intuitive for them. Other options I can think about just for brainstorming purposes:
  * `glu:`
  * `myst:`
* I kind like `paste` over output because I think it is more specific. calling it `nb:output` makes me assume it will be the output of a cell (which, one day, we may want to explicitly for cell outputs?).
* For the `nb:text` role, could this be `nb:paste` as well, with the assumption that by default, *anything* pasted in-line will be pasted as text?
  * To give myself a counter-point. I suppose (one day) we could try to add in-line options for things like in-line images (e.g., for [sparklines](https://en.wikipedia.org/wiki/Sparkline)) or for in-text HTML ([like tangle.js](http://worrydream.com/Tangle/))

**Comment by chrisjsewell** [comment]
> To give myself a counter-point. I suppose (one day) we could try to add in-line options for things like in-line images

Yeh that was my thinking, that this role (with the formatting aspect) is very specific to text

**Comment by choldgraf** [comment]
to further prove my counter point, I tried adding `` in-line to some parts of the docs, and this is what it looked like:

image

I think this means we can basically already include "sparkline"-style images easily. That is **super cool**.

**Comment by chrisjsewell** [comment]
Yeh maybe `glu` is more descriptive: `glu:text`, `glu:figure`. In this context `glu:paste` doesn't really make sense, so maybe `glu:object`?

**Comment by choldgraf** [comment]
@chrisjsewell yeah I was thinking that if we used `glu:` then we wouldn't need `paste`

**Comment by chrisjsewell** [comment]
> @chrisjsewell yeah I was thinking that if we used `glu:` then we wouldn't need `paste`

`nb:paste` -> `glu:?` (`glu:object`, `blu:block` ?)

**Comment by choldgraf** [comment]
yeah - I was thinking then we'd always follow a `glu:objecttype` pattern.

Or maybe `pst` for "paste"? We could register both `pst` and `paste` domains?

**Comment by chrisjsewell** [comment]
> We could register both pst and paste domains?

No I think only one domain name is allowed

**Comment by chrisjsewell** [comment]
Ok, for now, in 19b2b0e I have change the domain name to `glu` and changed `nb:paste` to `nb:any`.
I like `any` because it is reminiscent of the [sphinx `any` reference role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-any).

I have also added a `glu:any` (inline) role. This has required that I hack together an 'inline' version of `cell_output_to_nodes`: 
https://github.com/ExecutableBookProject/MyST-NB/blob/19b2b0edf0b4c686f4d21eb9a6d40c1b3e119f7f/myst_nb/transform.py#L64

Obviously this should be up-streamed to `jupyter-sphinx`.

See https://179-241268229-gh.circle-artifacts.com/0/html/use/glue.html#pasting-glued-variables-into-your-page 😄

File: myst_nb/nb_glue/domain.py [hunk 1]
```diff
@@ -0,0 +1,277 @@
...
+        return out_node
+
+
+class PasteInlineNode(PasteNode):
+    def create_node(self, output: dict):
+        """Create the output node, give the cell output."""
+        # the whole output chunk is deposited and rendered later
+        from myst_nb.parser import CellOutputBundleNode
+
+        output_node = CellOutputBundleNode(outputs=[output], inline=True)
```
choldgraf: I think here we could add an `inline` class, and then use that for styling?

myst_nb/nb_glue/domain.py
```diff
@@ -0,0 +1,277 @@
...
+    def key(self):
+        return self.attributes["key"]
+
+    def copy(self):
+        return self.__class__(location=self.location, **self.attributes)
+
+    def create_node(self, output: dict):
+        """Create the output node, give the cell output."""
+        # the whole output chunk is deposited and rendered later
+        # TODO move these nodes to separate module, to avoid cyclic imports
```
choldgraf: I hope that we can just import these nodes from jupyter-sphinx at some point, whenever I can figure out the tests in https://github.com/jupyter/jupyter-sphinx/pull/107

myst_nb/nb_glue/domain.py
chrisjsewell: Done in 306c406 (I arbitrarily added `max-width: 1.2em`) see: https://181-241268229-gh.circle-artifacts.com/0/html/use/glue.html#using-the-glu-any-role-directive

chrisjsewell: Somewhere in the documentation, we should probably add some instruction on modifying CSS styling

choldgraf: definitely - also, the inline images are awesome :-)

myst_nb/transform.py
```diff
@@ -49,3 +60,123 @@ def apply(self):
                 continue
             col = ImageCollector()
             col.process_doc(self.app, node)
+
+
+def cell_output_to_nodes_inline(cell, data_priority, write_stderr, dir, thebe_config):
```
choldgraf: This one you'd like to upstream yeah? What's the main difference w/ jupyter-sphinx one?

choldgraf:
This is looking really good to me - I like the API decisions in here, and we can always re-visit after others have had a chance to play with them.

myst_nb/transform.py
chrisjsewell: Yep. All the nodes returned are inline versions, i.e. `literal_block` -> `literal`, `container` -> `inline`,  `math_block` -> `math`

myst_nb/nb_glue/domain.py
```diff
@@ -0,0 +1,278 @@
...
+    """A sphinx domain for handling glue data """
+
+    name = "glu"
+    label = "NotebookGlue"
+    # data version, bump this when the format of self.data changes
+    data_version = 0.1
+    # data value for a fresh environment
+    # - cache is the mapping of all keys to outputs
+    # - docmap is the mapping of docnames to the set of keys it contains
+    # TODO storing all the outputs in the cache, could allow it to get very big
```
chrisjsewell: This should definitely be addressed at some point; if you end up 'glueing' 100s of plots, the cache is going to get very large, and most probably affect the load speed / memory usage of the sphinx environment

myst_nb/nb_glue/domain.py
```diff
@@ -0,0 +1,278 @@
...
+    # we may need to consider storing outputs on disc?
```
chrisjsewell: This should definitely be addressed at some point; if you end up 'glueing' 100s of plots, the cache is going to get very large, and most probably affect the load speed / memory usage of the sphinx environment

choldgraf: yeah for sure - maybe as a first step, we can add a check for items with type `image/png` in the notebooks and raise a warning once it gets over some arbitrary-but-high number?

chrisjsewell: Yeh all of that. Since we are already writing all these outputs to disc: https://github.com/ExecutableBookProject/MyST-NB/blob/1f88b44f88ec3051ca2d3cf5994a56980a377d78/myst_nb/parser.py#L151, it would be ideal to just link to those

choldgraf: Ah good point I think all we'd need to do is remove the data section from those outputs then

chrisjsewell: Ah yeh that's the one, fixed in c0a530c

choldgraf:
(FYI this is still in "draft" state but I am +1 to merge when you are happy with it)

chrisjsewell:
> (FYI this is still in "draft" state but I am +1 to merge when you are happy with it)

Two last changes for you to review above, both were made so that tables are formatted properly in figures here: see https://189-241268229-gh.circle-artifacts.com/0/html/use/glue.html

chrisjsewell:
If you do want to then merge, please add this as the commit message 😁 

This commit introduces the `NbGlueDomain` for controlling glue outputs.
The domain achieves two things:

1. It collates all roles/directives relating to rendering of 'glue' objects. So all you need to have in the extension setup is `app.add_domain(NbGlueDomain)`.
2. It centralises access to the glue cache, ensuring that keys related to deleted documents are removed

Multiple glue roles/directives have been introduced for formatting the output, including the `PasteFigure` directive, which allows for captions, labelling/referencing, width and alignment.

chrisjsewell:
Final piece of the puzzle: https://194-241268229-gh.circle-artifacts.com/0/html/use/glue.html#the-glu-math-directive

choldgraf:
This looks great to me - let's give it a merge and see how things feel as folks on the team use this, I am excited!

PR #74, opened 2020-03-14
adding simple binder link

choldgraf:
This PR will add simple functionality for Binder links with notebooks. Currently, all it tries to do is update the page context so that a template would have access to the Binder link. I'm not sure what is the best way to expose this in a way that *doesn't* use themes. I also wonder if `jupyter-sphinx` is also generating these links already with its thebelab support somehow, and if we could try piggy-backing on that? (@akhmerov do you have any thoughts on that?)


choldgraf:
closing as we now take care of this in sphinx-book-theme

PR #77, opened 2020-03-15
updates to glue docs

choldgraf:
This adds a few updates to the glue docs page as I was trying it out. It adds some more context about in-line images, tries to clean up the analysis code a bit so it is less-distracting, and clarifies a few points here and there.

Check out the page here: https://197-241268229-gh.circle-artifacts.com/0/html/use/glue.html

Check out the table at the bottom, pretty cool that it "just works"! https://197-241268229-gh.circle-artifacts.com/0/html/use/glue.html#pasting-into-tables


jstac:
This is brilliant.  I love this feature.  Great work and great docs.

Would `glu` be a better synonym for `glu:any` --- it seems more consistent...?

chrisjsewell:
Good doc improvements 👍 

> This is brilliant. I love this feature. Great work and great docs.
> 
> Would `glu` be a better synonym for `glu:any` --- it seems more consistent...?

Thanks @jstac, this is what I meant when I kept banging on about the Model-View pattern, it just makes life so much easier 😄

jstac:
I have seen the light :happy:

@najuzilu and @mtiley, please note these features.  As everything gets tied together, they can be included in the converted QuantEcon material.  More generally, the idea is to convert the material in order to replicate our output, as well as to exploit new features in order to make the output even better.

chrisjsewell:
Another use case that might be worth noting, is that now
you could have a notebook that doesn't actually show up in the text (you signify this to sphinx by adding `"orphan: True"` in the notebook metadata, see [file-wide-metadata](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#file-wide-metadata)), then you can have a separate markdown file that uses these roles/directives to show selected figures/data from one or more notebooks.

I've talked about this use case before; when you want to use the notebook(s) as a "log" of your data analysis, then have a separate document which is the final "publishable" output.

choldgraf:
I just added a commit that:

* Adds an orphaned notebook and pastes it into the `glue.ipynb` notebook
* Replaces `{glu:any}` with `{glu:}` (I assume that `{glu}` isn't possible? It didn't work for me)
* Slight reconfiguration

Note that currently it throws an "orphan" warning because we don't yet have the logic to parse notebook-level metadata in myst-nb (I think)

chrisjsewell:
> I assume that `{glu}` isn't possible? It didn't work for me

No, the `:` denotes to sphinx that the directive is part of a domain

PR #79, opened 2020-03-15
Add `glu:` alternative to `glue:any`

### LINKED ISSUE #78: Use `glue` as a placeholder for `glu:any`?
As I was typing the docs in #77 , I found it a bit tedious to keep typing `glu:any`. I wonder if we could use the full word `glue` for this particular use-case (and keep `glu:any` for more explicit usage).

So, for any output-specific command, use `glu:text`, `glu:figure`, etc. For generic outputs, use `glue` or `glu:any`.

What do folks think?

chrisjsewell:
closes #78 you can't use `glue`, as they have to be under the same domain, but you can use `glu:`

@jstac @choldgraf

### DIFF
--- a/myst_nb/nb_glue/domain.py
+++ b/myst_nb/nb_glue/domain.py
@@ -271,9 +271,9 @@ class NbGlueDomain(Domain):
     # - docmap is the mapping of docnames to the set of keys it contains
     initial_data = {"cache": {}, "docmap": {}}
 
-    directives = {"any": Paste, "figure": PasteFigure, "math": PasteMath}
+    directives = {"": Paste, "any": Paste, "figure": PasteFigure, "math": PasteMath}
 
-    roles = {"any": paste_any_role, "text": paste_text_role}
+    roles = {"": paste_any_role, "any": paste_any_role, "text": paste_text_role}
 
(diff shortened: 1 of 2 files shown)


choldgraf:
Amazing! I had assumed it wasn’t possible, which is why I was using “glue”. I’ll merge this in, and we can update the docs in #77

myst_nb/nb_glue/domain.py
```diff
@@ -271,9 +271,9 @@ class NbGlueDomain(Domain):
     # - docmap is the mapping of docnames to the set of keys it contains
     initial_data = {"cache": {}, "docmap": {}}
 
-    directives = {"any": Paste, "figure": PasteFigure, "math": PasteMath}
+    directives = {"": Paste, "any": Paste, "figure": PasteFigure, "math": PasteMath}
 
-    roles = {"any": paste_any_role, "text": paste_text_role}
+    roles = {"": paste_any_role, "any": paste_any_role, "text": paste_text_role}
```
chrisjsewell: @choldgraf yep, as you can see here, you can literally add as many pseudonyms for a role/directive as you want

Issue #12, opened 2020-02-18
A myst-based 'ipynb' document structure 


choldgraf:
Per a recent conversation with @mmcky and @chrisjsewell , we came up with the proposal for a MyST-based IPYNB structure.

**Here is an example notebook with the latest syntactic ideas**

````
---
kernel_info:
    name: python3
language_info:
    name: Python
title: "My notebook title"
comment: "If any of the above aren't specified then use jupyter defaults"
---

# Markdown syntax

## Cell breaks

We can manually break markdown cells quickly with this syntax

+++

### A subsection in another markdown cell

another proposal is to use

+++

### And here would be the other markdown cell...

## Markdown metadata

We can also explicitly separate a markdown cell and configure it like so:

```{markdown} tag1, tag2
---
key: val
---
## Here is some *configured* markdown!
```

And now this would be a third markdown cell

## Executable code

Code is always executed with 'execute' blocks, like so:

```{execute}
print('this would be run by the front-matter-specified, or default, kernel')
```

You can also add metadata to these

```{execute} kernelname
:key: val
:key2: val2
# Or perhaps we want a `metadata`: field for cell metadata, and other keys for options like jupyter-sphinx does
print('some python with cell metadata')
```
and that's it!
````
---
# influences

* Use MyST syntax to define code cells, markdown cells, and their breaks
* Don't *force* extra syntax when it's not needed. E.g., if there's pure markdown between two code cells, treat that as a markdown cell.
* Try to use markdown design influence in decisions, the syntax should suggest what it is doing.

# constraints

### Content within cells

All content withing cells, as well as the breaks between cells, should be 100% round-trippable in a lossless fashion. The markup language used in markdown cells and anything inside code cells should not be modified.

### Cell-level metadata

All metadata specified in markdown will be converted into the `ipynb` file. Conversely, a *subset* of cell-level `ipynb` metadata will be converted into markdown. **TODO: figure out what subset we want...only tags? Other publishing-specific stuff?**

### Notebook-level metadata

The same rule applies to notebook-level metdata (and we need to figure out the subset of metadata to keep)

# Proposed syntax

## Notebook-level metadata

A YAML header block at the top of the document will denote notebook-level metadata

example:

```
---
key: value
---

# My first header
```

## Code cells

[...]

PR #82, opened 2020-03-15
Add Myst-to-NB converter

chrisjsewell:
This PR introduces a function for converting text documents, written in the myst format, to a notebook. It could be used in the parser and in #55 to convert these documents to notebooks before executing/reading. See the test input/output for an example.

I would envisage this could be utilised by also adding `myst` to https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/parser.py#L27
This would denote files with the `.myst` extension as ones that should be parsed as notebooks, as opposed to normal `.md` files, then in https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/parser.py#L35
we would check the documents extension and apply this function, instead of `nbf.reads`, before proceeding with the rest of the process.

This addresses to a large extent #12


chrisjsewell:
Or we have a parser subclass:

```python
import nbformat as nbf
from myst_nb.parser import NotebookParser
from myst_nb.convert import myst_to_nb


class NotebookMystParser(NotebookParser):
    supported = ("myst",)

    def parse(self, inputstring, document):
        ntbk = myst_to_nb(inputstring)
        return super().parse(nbf.writes(ntbk), document)

```

chrisjsewell:
FYI there is an edge case that should probably be raised with sphinx: if you have multiple documents with the same name but different extensions; sphinx will not complain but will read the same one multiple times, e.g. if you had

```
a.md
a.rst
a.ipynb
a.myst
```

Only `a.rst` would be converted to `a.html`

choldgraf:
How do you see this usecase being used relative to writing in a `.md` file? Or more specifically - do you imagine users commonly writing three different kinds of documents: markdown, myst-notebook-markdown, and notebooks-with-myst-inside?

chrisjsewell:
Yes, I think these are three distinct document types. Although note, users could override this behaviour with [source_suffix](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source-suffix)

```python
source_suffix = {
    '.md': 'myst',
}
```

Which would make all `.md` be read "as notebooks"

chrisjsewell:
Note this kind of an alternative to using the `jupyter-sphinx` directives; the problem with them being (a) at present you end up with different formatting between the two, but mainly (b) how can they be tied in to #55; where we want to identify all the notebooks to execute, before any parsing has been done

chrisjsewell:
Perhaps the default extension for these files could be `.mystnb`, and this is also what jupytext would convert to/from

chrisjsewell:
In 5d50424 I have added the full round-trip conversion. `nb_to_myst` could probably be added directly to jupytext, but `myst_to_nb` uses `myst-parser` so they probably wouldn't want that as a dependency and would want to re-write.

choldgraf:
Actually, this comment: https://github.com/mwouts/jupytext/issues/447#issuecomment-596464369 makes me think that marc would be happy for us to make myst a requirement for the ntbk <--> myst markdown sync. When people want to use pandoc markdown he just calls pandoc, and myst-parser is a more lightweight dependency than pandoc.

Maybe @AakashGfude could give a look at @chrisjsewell 's code and see how tough it'd be to use that as the basis of a PR to jupytext?

choldgraf:
just to confirm - if the jupytext PR gets merged, then that will supercede this one right?

choldgraf:
(and, just another thought in case it seems relevant in the other PR, we could also keep the merging logic in this repository and have Jupytext use a function that myst_nb provides, or something like this...that might give us more iterative control over the code without needing to wait for upstream reviews in jupytext)

chrisjsewell:
> just to confirm - if the jupytext PR gets merged, then that will supercede this one right?

It would yes

> (and, just another thought in case it seems relevant in the other PR, we could also keep the merging logic in this repository and have Jupytext use a function that myst_nb provides, or something like this...that might give us more iterative control over the code without needing to wait for upstream reviews in jupytext)

The problem with this is that all the testing is in jupytext. So any changes to the function might break these tests, but you wouldn't know until you tried to use the function in jupytext

choldgraf:
@chrisjsewell makes sense - and to be clear I *prefer* that this code is an upstream contribution to jupytext, I'm just a bit worried if we need to make some rapid changes to the codebase etc.

choldgraf:
also let's discuss the `{execute}` vs `{nb-code}` thing here, so it doesn't clutter the other PR (since Marc has no context for that conversation).

I added `{execute}` as a synonym for `{jupyter-execute}` only as a stopgap in this repository until we could get a proper "myst-based notebook format" ready to go. I think that asking users to write `{execute}` is a more intuitive than asking them to write `{nb-cell}`, so my preference is to use that pattern, rather than to *only* use `{execute}` as a short-hand for `{jupyter-execute}`

choldgraf:
(happy to hop on a phone call to discuss this if that's a faster way to resolve this. I think that the "notebook format in myst" should match *exactly* what we have in the issue in this repository, unless the group decides otherwise, since that's what everybody agreed on the last time we discussed)

chrisjsewell:
>  I think that asking users to write `{execute}` is more intuitive than asking them to write `{nb-code}`

I wouldn't say I have any strong preference either way TBH.

Firstly though, I wanted to clarify the distinction, and note that if we do use `execute`, then we should remove https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/__init__.py#L99

Secondly, from a different perspective, if you're thinking of the file in terms of a notebook, isn't it more intuitive to use `nb-code` which implies directly that this is a notebook code cell?

Thirdly, `{raw}` is already a sphinx directive, so we would have to use a different name.

chrisjsewell:
Also, it does feel a bit of a "fudge" that these directives aren't technically directives (i.e. that can be directly used by sphinx), but I don't see any better trade-off, to achieve a full round-trip capability?

choldgraf:
> Firstly though, I wanted to clarify the distinction, and note that if we do use execute, then we should remove

Yep, that was my intention all along. I just added the `add_directive` bit to give us something to work with for development purposes.

> Secondly, from a different perspective, if you're thinking of the file in terms of a notebook, isn't it more intuitive to use nb-code which implies directly that this is a notebook code cell?

> Thirdly, {raw} is already a sphinx directive, so we would have to use a different name.

I think these are both reasonable points - but the point is that we need to decide as a group what the syntax should be, we can't choose different names just at the time of a PR without getting feedback from others. I think these are reasonable options, and you make a good point about clobbering things like `{raw}`, but especially for things like user-facing API of the core myst-markdown format, others should have an opportunity to weigh in.

Another option is to treat this as a stop-gap for the sake of making progress, let marc know that the names may change but we're happy to make a PR to do this if it happens, then update [this issue](https://github.com/ExecutableBookProject/MyST-NB/issues/12) to have further discussion about how the proposed format should change.

WDYT?

chrisjsewell:
> Yep, that was my intention all along. I just added the add_directive bit to give us something to work with for development purposes.

👍 

> but especially for things like user-facing API of the core myst-markdown format, others should have an opportunity to weigh in

Oh yeh absolutely. I have actually written these functions to take the directive names as variables: 
https://github.com/ExecutableBookProject/jupytext/blob/8999b1d93ee4938ddbc36fd749c02639b5d9b588/jupytext/myst.py#L98-L100
with the original intention in this PR of leaving that decision until later.
But then for the jupytext PR obviously they have to be solidified

> Another option is to treat this as a stop-gap for the sake of making progress, let marc know that the names may change but we're happy to make a PR to do this if it happens, then update this issue to have further discussion about how the proposed format should change.

yes I'd be happy with the route

chrisjsewell:
> then update this issue to have further discussion about how the proposed format should change.

The good thing about having now made this initial jupytext implementation, is that we have a whole suite of output `.mystnb`/`.mnb` files, for everyone to inspect and get a feel for how they are looking:

https://github.com/ExecutableBookProject/jupytext/tree/mystnb/tests/notebooks/mirror/ipynb_to_myst

choldgraf:
@mmcky and @jstac would be great to hear your thoughts

jstac:
The syntax here looks perfectly fine to my eyes: https://github.com/ExecutableBookProject/jupytext/blob/mystnb/tests/notebooks/mirror/ipynb_to_myst/Notebook%20with%20metadata%20and%20long%20cells.mystnb

I don't have strong opinions on this point.

jstac:
Kind of separate, kind of related, could someone please clarify:

So we now have `myst` files and `mystnb` files.  Which is the end user supposed to author in?

My understanding was that they would author in myst, as in [this example](https://raw.githubusercontent.com/ExecutableBookProject/MyST-Parser/master/docs/examples/wealth_dynamics_md.md)

But if they want to use live sync with jupytext, then they must read and write in mystnb --- right?  Is this, in the end, the text-based source format we refer users to?  Is that the one they should learn and interact with?

Because they shouldn't have to be familiar with both...

chrisjsewell:
@jstac copied from https://github.com/mwouts/jupytext/pull/456#issuecomment-599634947

With the `.mystnb`/`.mnd` file format, we are saying that it will not be treated the same as simple `.md` in the sphinx conversion; it will be specifically converted to a notebook first (either using a pre-conversion step in [NotebookParser](https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/parser.py#L24) or a separate sub-class of it [as I outlined here](https://github.com/ExecutableBookProject/MyST-NB/pull/82#issuecomment-599254868)). So in your sphinx project you may have three distinct types of file:

- `.md` which are read directly by the myst-parser package's [MystParser](https://github.com/ExecutableBookProject/MyST-Parser/blob/62e1bf9a95fa683b4c4e2aaac1ad4fd94b855ce6/myst_parser/sphinx_parser.py#L12). These could contain `jupyter-execute` directives (from `jupyter-sphinx`), **but would not** be involved in any jupyter-cache like execution caching, invoked by myst-nb.
- `.ipynb` which are read by the myst-nb package's [NotebookParser](https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/parser.py#L24)
- `.mystnb` which are read by `NotebookParser` or a sub-class like `MystNbParser`.

In this context, `.ipynb` and `.mystnb` are interchangeable, you could choose to use one or the other and just switch between them using jupytext. (note: we would have to work out how to make sphinx ignore one if both were present)

So the long and the short of it is: any documents that have no executable code in: you would just use `.md`, and any with code, you would use `.mystnb`. The syntax is exactly the same in both these files (i.e. MyST) but the `.mystnb` has the 'special' notebook elements (`{nb-code}`, `{nb-raw}` and `+++` to denote the start of a new markdown cells).

And to hammer home the reason for having this separate `mystnb` extension is that it makes it easy to rapidly identify which files are 'notebook-like', and quickly convert them to notebooks, so that we can implement things like initial parallel execution (see #55)

chrisjsewell:
> My understanding was that they would author in myst, as in this example

Note to 'convert' this example it would mainly just involve find/replace of `{code-block}` to `{nb-code}`

and this:

````
```{code-block} ipython
---
class: hide-output
---
!pip install --upgrade quantecon
```
`````

would become this:

````
```{nb-code} ipython
---
tags: [hide-output]
---
!pip install --upgrade quantecon
```
````

or shorter:

````
```{nb-code} ipython
:tags: [hide-output]

!pip install --upgrade quantecon
```
````

**Comment by choldgraf** [comment]
Just a note that I think the biggest downside here is that we'd be asking users to keep track of four different kinds of files in their head (`.ipynb`, `.md`, `.mystnb`, `.myst`), two of which are novel (`.myst`, `.mystnb`). In my mind, we were just going to keep this as "myst flavor" of markdown, and use `.md` so users didn't have to change. In the same way that GitHub-flavored markdown, kramdown markdown, commonmark, etc all use the same `.md` extension, this would be another.

I agree that it makes the developer burden higher because then you have to check what kind of file it is first...but the other option is to make the user burden higher :-/

**Comment by jstac** [comment]
Following @choldgraf 's point, I think this will cause significant confusion.

I think we should be very clear about the text source format that the average user will interact with.  If we're going to round-trip with ipynb files --- which has always been on the agenda --- then users are inevitably going to have to interact with `mystnb`, aren't they?  

For example, if they use jupytext to author and keep their files in sync?

I that case I don't think we should also ask them to be familiar with myst itself.  I realize the differences are small, but this small deviation can be even worse in terms of confusion.

**Comment by chrisjsewell** [comment]
> In my mind, we were just going to keep this as "myst flavor" of markdown, and use .md so users didn't have to change. In the same way that GitHub-flavored markdown, kramdown markdown, commonmark, etc all use the same .md extension, this would be another.

Well yes you can certainly just use `.md` for *non-notebook* files. A possible reason for using `.myst` would be if you want to have it specifically identified as MyST flavour, for future syntax highlighters and LSP, but is definitely not a necessity.

For the distinction between `.md` and  `.mnb`, well the counterpoint here is RMarkdown, they don't just use `.md`, I think for good reason. Plus, with advancements in jupytext, in essence you should never need to directly use `.ipynb` as you can just open `.mnb` directly in jupyter notebook/lab.

**Comment by chrisjsewell** [comment]
Its certainly something we can play around with though, to work out what is the best compromise

**Comment by choldgraf** [comment]
But a "notebook written in myst" file is still valid myst, right? That makes me think we should just keep them with the same extension to avoid confusing users. If `mnb` is just "a particular configuration of valid myst" it doesn't seem necessary to require users to remember a whole new file extension type, no?

> well the counterpoint here is RMarkdown

If MyST ever becomes remotely as popular as RMarkdown, then I think it is fine if we have a separate extension for it haha, in the meantime, I worry that people will be confused. The issue is in the "N+1 format" perception of this, adding in the fact that most users will not know anything about this, and will probably be low-motivation to actually try it out, so we want those barriers to be as small as possible.

**Comment by chrisjsewell** [comment]
> But a "notebook written in myst" file is still valid myst, right? 

Errr, well yes and no, the question is how would you modify jupyter-sphinx (or some other sphinx extension) to use this:

````md
---
kernelspec:
  display_name: Python 3
  language: python
  name: python3
---

```{nb-code} ipython3
---
other:
  more: true
tags: [hide-output, show-input]
---

print("Hallo!")
```
````

The answer is, you can't; they work in fundamentally different ways.

So if you do just call every file `.md`, you can't just use the `myst-parser` extension, (and there is certainly no point using `jupyter-sphinx`); you will need to write in additional steps/phases to the `myst-nb` extension to "pre-parse" all these files, and decide whether they are "notebook-like" or not, and treat them accordingly.

This can be achieved, but there would certainly be an overhead in terms of slow overall parsing and some added complexity. It whether you are willing to live with that.

**Comment by mmcky** [comment]
hey guys. sorry have had my head in the split of the `python` lecture sets. My 2c.
First @chrisjsewell nice work! this will be wonderful. Like you say we can now generate `myst` documents from `ipynb` files! Hooray. 

**Comments:**

1. I think it is important to only use `.md` and `.myst`. I would use the later as a convenience reminder of what is in the file for `text-based format`
1. Similary I think we should use `.ipynb` or `.mystnb` (or some variant) for the `machine-based format`

I think if at all possible we use figure out a `single` syntax that is `two-way` compatible so we don't have `{execute}` and `{code-nb}`. I don't think the markup should have to cater to the notebook structure so explicitly and I think will find it confusing when to use which. 

I haven't fully comprehended the need to have both `execute` and `code-nb` for round-tripping support from a technical perspective -- which I know is important. @chrisjsewell I don't understand why the meta (`yaml` or `short-hand` syntax) couldn't be added to cell metadata? Sorry if you feel like you've explained this already.

**Comment by choldgraf** [comment]
The way I imagined it, is that `myst-nb` would define its own directives (either `{execute}` or `{nb-code}` etc). In the future it wouldn't know anything about `jupyter-sphinx` (beyond maybe re-using some of jupyter-sphinx's tools for its own directive). So, you could write something like this:

````
---
kernelspec:
  display_name: Python 3
  language: python
  name: python3
---

```{nb-code} ipython3
---
other:
  more: true
tags: [hide-output, show-input]
---

print("Hallo!")
```

And here in the markdown users could call directives:

```{jupyter-execute}
print("jupyter-sphinx will execute this")
print("but it knows nothing about myst-nb etc, it's just another directive
```

````

**Comment by chrisjsewell** [comment]
I feel we should just scrap `{execute}`: because, byt MyST-NB's very nature (and for round-tripping) we are dictating that files that are executable adhere to a "notebook-like" structure (i.e. one kernel per file, and code cells must be top-level / not wrapped in any other directives, etc). Personally I feel it is a lot easy to be able to think about these executable files in relation to notebooks; having `{nb-code}` rather than `{execute}`. Do we expect our users will not have used notebooks before?

**Comment by choldgraf** [comment]
I'm fine w/ `{nb-code}` 👍 

What if we used a domain for this? Could we have something like `{nb:code}`, `{nb:markdown}`, etc? And then we could even have some helpers like `{nb:python}` that would pick available python kernels etc?

**Comment by mmcky** [comment]
I guess I don't understand why it can't just be `{code}` to identify the contents of the directive as `code` directly rather than link these constructs to the `notebook` terminology?

**Comment by chrisjsewell** [comment]
> The way I imagined it, is that myst-nb would define its own directives

In your example

- how does this extension obtain the kernel data from the front-matter? (bearing in mind that rST does not have front matter)
- what happens if the user adds extra metadata to the code-cell and then round-trips?

All I'm saying you, you go and create a PR for this sphinx extension, and get it to work, and I'll be happy to eat my hat. But, until then, I'm saying its not possible (and certainly not something that I'm going to try to write 😜 )

**Comment by chrisjsewell** [comment]
> I guess I don't understand why it can't just be {code} to identify the contents of the directive as code directly rather than link these constructs to the notebook terminology?

because code is already a directive used by sphinx/docutils to signify "highlighted static code"

**Comment by chrisjsewell** [comment]
> What if we used a domain for this? Could we have something like {nb:code}, {nb:markdown}, etc?

`{nb:code}` and `{nb:raw} would be fine, `{nb:markdown}` isn't a thing because markdown is never wrapped

**Comment by chrisjsewell** [comment]
> And then we could even have some helpers like `{nb:python}` that would pick available python kernels etc?

This wouldn't work. I think you loved by `glu:` domain so much, you are trying to apply it to everything now 😜 😆

**Comment by mmcky** [comment]
>> I guess I don't understand why it can't just be {code} to identify the contents of the directive as code directly rather than link these constructs to the notebook terminology?

> because code is already a directive used by sphinx/docutils to signify "highlighted static code"

I see. I guess in the `markdown` context I was thinking that 
````
```<lang>
```
````
would translate to the `static` concept, where as:

````
```{code}
```
````

fulfilled the role of execution. In our `sphinxcontrib-jupyter` approach we intercept all `code-blocks` (+ `literal_include etc`) and add them to the notebook for execution. If someone wants a static code block they would use `:no-execute: True`. But overriding `code` in `docutils` I guess would cause complications.

**Comment by chrisjsewell** [comment]
> If someone wants a static code block they would use :no-execute: True. But overriding code in docutils I guess would cause complications.

Yeh I just think that's super confusing; to have some {code} directives that do execute and some that don't, and also this doesn't really track with round-trip conversion to notebooks.

Also

````
```<lang>
```
````

Are converted to `code-block` in the dicutils AST, so `sphinxcontrib-jupyter` would just intercept them as well, which would be even more confusing!

**Comment by choldgraf** [comment]
Mostly I am trying to not deviate as much as possible from RMarkdown's patterns of `{r}`and `{python}`...ideally users would be able to just use that exactly in order to define executable code, since it's such a common pattern already. I'm just not sure if that would be possible without colliding with other things in Sphinx, which is why I was thinking `{nb:python}` etc.

**Comment by mmcky** [comment]
I see. I had assumed that the parser would essentially interpret:

````md
```<lang>
```
````
as
````rst
.. code-block::
   :no-execute: True
````

but the `:no-execute:` isn't currently part of `docutils` option set for that directive. Something we would have to add. I guess the only difference is that you're not using `metadata` to control `execution` but rather using a different directive. That makes sense and happy with that. 

From my viewpoint I don't think `{nb-}` is a good name because I think there will be a cohort of users who just write in `myst`. Despite having an exact counterpart in `notebook` authors shouldn't have to think in notebooks.

**Comment by chrisjsewell** [comment]
> authors shouldn't have to think in notebooks.

> Mostly I am trying to not deviate as much as possible from RMarkdown's patterns of {r}and {python}...ideally users would be able to just use that exactly in order to define executable code, since it's such a common pattern already. I'm just not sure if that would be possible without colliding with other things in Sphinx, which is why I was thinking {nb:python} etc.

The common pattern here is jupyter notebooks, and how they run code, not however RMarkdown runs code; we, and `jupyter-sphinx`, and `sphinxcontrib-jupyter`) are all literally converting the blocks to notebooks and running them. I think we should absolutely be saying to authors "think of the code blocks running in terms of notebooks", that is the simplest way for them to understand the execution: one kernel per notebook, that dictates the language/executable that all of the cells in that notebook are run with. 

This is why `{r}` or `{python}` or `{nb:python}` makes no sense to me, because there is no guarantee that that is the language that the code cells will actually be run with. You are just going to have multiple names for exactly the same underlying directive. We would also have to have one of these directive names for at least every one of the languages supported by jupytext:

```python
_JUPYTER_LANGUAGES = ['R', 'bash', 'sh', 'python', 'python2', 'python3', 'javascript', 'js', 'perl',
                      'html', 'latex', 'markdown', 'pypy', 'ruby', 'script', 'svg',
                      'matlab', 'octave', 'idl', 'robotframework', 'spark', 'sql']
```

jstac:
I agree.  Jupyter notebooks is the right mental model for authors, one kernel per notebook.  I imagine most people will start off authoring in ipynb anyway, and then gradually start to use text based source as they seek extra features.

mmcky:
> Jupyter notebooks is the right mental model for authors, one kernel per notebook.

I agree that thinking in terms of notebooks is a good way to think about how the blocks are executed but I don't think the syntax should be tied to `notebooks`. It is ok if I get outvoted on this but I guess my point is when I am writing a myst text file I wouldn't think to naturally write `{nb-code}` for an executable code block as I am writing a document not a notebook.

chrisjsewell:
> when I am writing a myst text file I wouldn't think to naturally write `{nb-code} `

But the converse is, if I had converted from a notebook then seeing `{nb-code}` would be pretty intuitive

choldgraf:
Personally, I agree with you both in terms of `{nb-code}` or `{execute}` being better names than `{python}` or `{r}`. I'm just thinking in terms of the behemoth in this space, which is RMarkdown, and our opportunity to bring some of those folks onto this stack by making it as painless as possible.

I'd be fine with either `{nb-code}` or `{execute}/{exec}`....hmmm

what about `{cell}` or `{code-cell}`? Those are not quite "notebook-specific" but they convey the same idea. If you're coming from a notebook you'll understand what they mean, but if you aren't then they still make sense?

chrisjsewell:
> . I'm just thinking in terms of the behemoth in this space, which is RMarkdown, and our opportunity to bring some of those folks onto this stack by making it as painless as possible.

Yeh I absolutely take your point, its just trying to get the balance right. Remember that, since jupytext now round-trips both RMarkdown and MyST-NB, we do have a ready-made conversion tool from RMarkdown 😄 

`{code-cell}` would be fine by me, not `cell` because we would also have `{raw-cell}`

mmcky:
I think `{code-cell}` would make the syntax clearer.

chrisjsewell:
See  mwouts/jupytext@d37f73a

choldgraf:
thanks for the quick discussion and turnaround on this one all, much appreciated!

it definitely does make me more responsive when I'm not supposed to leave the house 😆

jstac:
I hear what you're saying about Rmarkdown, @choldgraf , although perhaps it's possible to overstate the number of would-be authors who are coming from that environment.  I imagine that the great majority will be notebook users hoping to soup up the presentation of their results, while knowing little about Rmarkdown.

(Like me :grin:)

I suspect most (but not all) Rmarkdown users are R users, who have good tools already in terms of EBP substitutes.  

Anyways, +1 for `{code-cell}`.

choldgraf:
Hey all - another decision-point has come up in @chrisjsewell's PR to jupytext:

Should the extension coming out of jupytext be `.md` or `.mnb` or `.myst.md`?

My preference as you can guess is for plain old `.md`, I think it will confuse casual users (AKA, most of our users) if there is anything other than `.md` in their file extensions. But, I am open to alternative points. Chris made the good point that having a myst-specific extension would make it much easier to know when a markdown file is "myst markdown". That said, the more we make people do anything different from what they're doing, the more people will not try out this stack in the first place.

jstac:
Looks like you got your preference: https://github.com/mwouts/jupytext/commit/62351f45a843f0c877dcc9bfdc6a33016cc388fb

chrisjsewell:
> Looks like you got your preference: [mwouts/jupytext@62351f4](https://github.com/mwouts/jupytext/commit/62351f45a843f0c877dcc9bfdc6a33016cc388fb)

Yeh I can't be bothered to argue about it lol. It does mean that I'll probably put the LSP on the back-burner though. Because, if there is no way to distinguish myst from md, then it doesn't really work.

chrisjsewell:
Also these files won't work with the `myst-parser` extension. You need to add a separate parser to this package that calls [`jupytext.myst.matches_mystnb`](https://github.com/ExecutableBookProject/jupytext/blob/3263403fce77fd760f61fc875c96f1a82ed79ad9/jupytext/myst.py#L38), to work out if the markdown file should be read as pure markdown or converted to a notebook.

@AakashGfude will also need to use it in some manner in #55 to work out which markdown files to convert/execute/cache

jstac:
> Yeh I can't be bothered to argue about it lol.  It does mean that I'll probably put the LSP on the back-burner though. Because, if there is no way to distinguish myst from md, then it doesn't really work.

In that case I can definitely be bothered to argue about it.

In terms of the user interface, this is hugely important.  I can tell you this as someone who has been authoring in jupinx for a long time.  The lack of editor support for the source format is absolutely the first thing I would change if I could.

I think user confusion between `md` and `mnb` is likely to be minor.  In fact, as a user rather than a developer, I would strongly prefer `mnb` because this is a one-to-one mapping with a notebook.  It is, from the very start, a markdown representation of a notebook.  This is exactly the mental model that we want users to have in mind.  So `mnb` is perfect.

jstac:
Sorry I'm late to the party on this --- clearly I didn't understand all the implications --- but should we put a hold on https://github.com/mwouts/jupytext/pull/458 until this is decided?

choldgraf:
Given what @chrisjsewell mentioned I think I'd be fine if we used `.myst` or `.mnb`. We can always revisit this later if it seems like a big problem. Would it be enough for our usecases?

I am also happy to hop on a call to discuss this if that's helpful to hash out the details.

chrisjsewell:
> Sorry I'm late to the party on this --- clearly I didn't understand all the implications --- but should we put a hold on [mwouts/jupytext#458](https://github.com/mwouts/jupytext/pull/458) until this is decided?

No, that's probably not necessary. Note that, with the current PR state, if you have jupytext installed and this at the top of your file:

```markdown
---
jupytext:
  text_representation:
    extension: .myst
    format_name: myst
    format_version: '0.7'
    jupytext_version: 1.4.0+dev
kernelspec:
  display_name: Python 3
  language: python
  name: python3
---
```

Then any of the following extensions will open as notebooks:

image

chrisjsewell:
I think theres a way forward on this, so mwouts/jupytext#458 can definitely be merged. Then @choldgraf can decide if you want to work out how to make the default pairing (if going from `ipynb` -> text based) be to `.md`, rather than default now which is `.mnb` (see https://github.com/mwouts/jupytext/pull/458#issuecomment-600239217)

choldgraf:
Let's keep it mnb for now, and see how things go.

chrisjsewell:
> Let's keep it mnb for now, and see how things go.

Yeh cool cool, if you mention that to mwouts in the PR then, and hopefully we can get a jupytext release with this code in soon, to utilise with https://github.com/ExecutableBookProject/MyST-NB/pull/82#issuecomment-600251760

choldgraf:
done 👍

chrisjsewell:
Cheers 😁

jstac:
Thanks all!!

mmcky:
Nice work! Can I just say it's great working in a team that is open to debate and dialogue on these issues. The project will be better off for it.

chrisjsewell:
MyST format is now merged into jupytext 🎉 https://github.com/mwouts/jupytext/releases/tag/1.4.1

PR #89, opened 2020-03-19
myst-parser version

choldgraf:


setup.py
```diff
@@ -41,7 +41,7 @@
     python_requires=">=3.6",
     package_data={"myst_nb": ["_static/mystnb.css"]},
     install_requires=[
-        "myst-parser~=0.6.0",
+        "myst-parser~=0.7.0",
```
chrisjsewell: Use 7.1 I think (there was an addition for the Jupytext functionality

chrisjsewell:
You need to also do the pandas theme updates

choldgraf:
good point, may as well do it here too

choldgraf:
latest push uses the pydata theme, let's see if the docs are happy

PR #91, opened 2020-03-20
updating glu to glue in directive+role

choldgraf:
This updates `glu:` to `glue:`. This just bit me in a live demo, so I wanna switch it over sooner than later.

@chrisjsewell are OK with this switch? After trying to do it with a bunch of people watching, my first intuition was definitely to use the full word "glue"...


choldgraf:
I haven't heard any protests so I'm gonna assume folks are +1 on this and give it a merge

PR #92, opened 2020-03-21
pydata-sphinx namechange

phaustin:
update dev instructions


choldgraf:
Good catch, thanks very much!

PR #100, opened 2020-03-25
udpating docs so bokeh and plotly show up

choldgraf:


choldgraf:
This updates the interactive page so that some of the plots should show up better 👍 

getting interactive content to work in outputs will be tricky, I think there is still more to do because it is highly library-specific

PR #102, opened 2020-03-27
Add equation example to basic.ipynb example jupyter notebook

mtiley:
@jstac from the discussion here https://github.com/ExecutableBookProject/meta/issues/45

I don't think the changed python version in the metadata will be a problem but I can revert it to 3.8.0 if you like and update this PR


jstac:
This works nicely, thanks.

choldgraf:
excellent, thanks!

PR #104, opened 2020-03-28
minor edits to docs

jstac:
Just some small edits to the documentation on execution and jupyter-cache.

PR #105, opened 2020-03-28
Add development documentation

chrisjsewell:
Add a page on package development, which @jstac can use to fix he's PR #104 😆 

https://267-241268229-gh.circle-artifacts.com/0/html/develop/contributing.html

Also @AakashGfude, perhaps you would like to add a page about writing/running tests, as you have a pretty good understanding of that now 😁

### DIFF
--- a/docs/develop/contributing.md
+++ b/docs/develop/contributing.md
@@ -51,10 +51,6 @@ For documentation build tests:
 >> make html-strict
 ```
 
-```{seealso}
-{ref}`develop/testing`
-```
-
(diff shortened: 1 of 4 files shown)


choldgraf:
This LGTM - I think we should just give it a merge and @AakashGfude it would be awesome if made a follow-up PR with some testing instructions :-)

AakashGfude:
@choldgraf @chrisjsewell Gotcha. I will work on this tomorrow.

Issue #71, opened 2020-03-14
Fix line number / cell index reporting


chrisjsewell:
So that sphinx will report errors/warnings referring to correct place in the notebook

choldgraf:
Could we pass in the cell number into the state machine somehow? I feel like cell numbers are more helpful to report here than line numbers

chrisjsewell:
Well ideally you'd report both. Its just thinking of a way to do this that ideally still keeps `myst-parser` (where some of this logic would need to be added) agnostic to notebooks.

Note the cell index is now stored in the markdown AST: https://github.com/ExecutableBookProject/MyST-NB/blob/4c3f36b884086878dc0af5dfcfd735b98ebbbacf/myst_nb/parser.py#L94,
its just then how to get this into the docutils AST and reported by the docutils reporter.

PR #107, opened 2020-03-29
Move to markdown-it-py markdown parser implementation

chrisjsewell:
Using an alpha release of ExecutableBookProject/MyST-Parser#123

All the current tests pass, with no changes to the expected outputs. I also took the opportunity to split up the code in `NotebookParser.parse` in to separate functions.


chrisjsewell:
From a quick scan of the documentation, I don't see any issues.

However, I do note that the plotly and ipywidgets no longer show, moving from:

- https://myst-nb.readthedocs.io/en/stable/use/interactive.html
- https://myst-nb.readthedocs.io/en/latest/use/interactive.html

I assume this is because this notebook is now being executed during the build, and is thus encountering  jupyter/nbclient#39

chrisjsewell:
In 74cf6a3 I have introduced some nice, self-documenting render tests, see tests/nb_fixtures/basic.txt

chrisjsewell:
In 536a1dd I have added a simple fix for cell index / line number reporting:

Line number are now set as `<cell index>*10000 + <line number>` (I'm guessing a cell won't have more than 10000 lines!). This is a simple solution to addresses #71, that doesn't require any complex overrides of the sphinx reporting machinery.  We should also document this.

choldgraf:
I think this looks great, and agreed about the widgets not showing up. I don't think that should block us here

PR #108, opened 2020-03-29
Fix cache outputs retrieval for invalidated notebooks

chrisjsewell:
The previous implementation fails when an existing notebook is changed with failing code.

For example make the MyST-NB documentation, then add `raise Exception` to a cell in `use/basic.ipynb`, and re-run:

image

With the new (simplified) code:

image

@AakashGfude is there any reason why you were trying to find the cache record by URI? (something you really shouldn't be trying to do 😬)

### DIFF
--- a/myst_nb/cache.py
+++ b/myst_nb/cache.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from sphinx.util import logging
 from sphinx.util.osutil import ensuredir
 
-from jupyter_cache.cache.db import NbCacheRecord
 from jupyter_cache import get_cache
 from jupyter_cache.executors import load_executor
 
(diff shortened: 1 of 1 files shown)


choldgraf:
This all seems good to me - I'll hold off on merging in case @AakashGfude had a specific reason for the old pattern

AakashGfude:
@chrisjsewell  I am actually trying to get the list of all records with that URI and then iterating through them to find the latest record. Reason being, I only have uri as an identifier at that point in the function. 
```
cache_list = NbCacheRecord.records_from_uri(file_path, db)
        if len(cache_list):
            latest = None
            for item in cache_list:
                if latest is None or (latest < item.created):
                    latest = item.created
                    latest_record = item
            cache_record = latest_record
```

choldgraf:
Let's merge away then!

PR #109, opened 2020-03-30
Cache: get source suffixes from sphinx app, rather than hard-codind to `.ipynb`

chrisjsewell:


chrisjsewell:
> I guess this is setting the stage for MD execution yeah?

Yeh, plus just making it more robust with the proper sphinx machinery

PR #112, opened 2020-03-31
Documentation for unit testing

AakashGfude:
@choldgraf @chrisjsewell, have added a few lines on unit testing. What do you think about this? Should I make it more/less verbose? Should I also write about using `mock_document`, `MockEnv`?

### DIFF
--- a/docs/develop/contributing.md
+++ b/docs/develop/contributing.md
@@ -76,3 +76,36 @@ Merging pull requests: There are three ways of 'merging' pull requests on GitHub
 [rtd-link]: https://myst-parser.readthedocs.io/en/latest/?badge=latest
 [black-badge]: https://img.shields.io/badge/code%20style-black-000000.svg
 [black-link]: https://github.com/ambv/black
+
+## Unit Testing
+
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Should I also write about using `mock_document`, `MockEnv`?

FYI in #107 I have removed these, and the tests now run as actual sandboxed sphinx builds. You might be interested to check it out.
This doesn't change what you have already written though, which looks fine 👍

choldgraf:
nice! agreed that this is an improvement. @chrisjsewell if you think we're moving away from `MockFoo`, then I think we can leave it out of the docs. As an aside, I think it's a simpler approach *not* so use these MockEnvs...it feels a little bit brittle since we aren't testing in the *actual* environment that Sphinx has when it runs.

chrisjsewell:
Well there’s a hierarchy of testing; where possible there should be low-level isolated tests per function, then other full scale integration tests. 
But yeh, for this package the MockEnv was getting to complex, so has been removed.

choldgraf:
@chrisjsewell I agree - I just meant that when you're close to Mocking the entirety of a sphinx build, it makes sense to just use the whole sphinx build 😆

PR #116, opened 2020-03-31
Draft of text-based notebooks

chrisjsewell:
Builds on top of #107. In this implementation `myst-nb` replaces `myst-parser` as the markdown parser and tests all files for a "notebook signature".

Documented in https://306-241268229-gh.circle-artifacts.com/0/html/use/markdown.html


chrisjsewell:
Some thought needs to be put into how people might use this with jupytext, i.e. if there is both a `notebook.ipynb` and `notebook.md`  in the source folder. At the moment its rather undetermined which one will be used to actually build the document with. I raised this in sphinx-doc/sphinx#7324, but they only went half-way with my suggestion; such that a warning will now be raised, but really you should be able to select a preferential extension.

choldgraf:
Fantastic - I think this is one of the final pieces in the puzzle!

re: extension preference, this is something that we could also control at the `jupyter-book` level. Basically we could do a quick check for any files with duplicate filenames, and for any that are found, we could load one of them into `exclude_patterns` before passing the whole thing to Sphinx

jstac:
Great work @chrisjsewell.

(Regarding the source-for-build policy at the jupyter-book level, it is a potential source of confusion isn't it?  Given that most people are pretty careless when it comes to reading the documentation.   Perhaps it's worth having a clear statement of the default (either md or ipynb), although still allowing the user to customize.  A natural default would be to build from the md whenever they exist.  That will never trouble people who use only ipynb.)

chrisjsewell:
The correct line numbers are now reported for these text-based notebooks 😄 

Shall I also convert some more of the notebooks in the documentation to markdown files? Obviously the markdown ones are a lot easy to edit (and better for git diffs), but probably also want some ipynb in their for example.

choldgraf:
I'm +1 for the idea. Let's leave at least one ipynb file in there, but I'm fine with almost all of them being `.md`.

choldgraf:
@jstac I agree - my intuition is that "markdown should take precedence by default". That way, we won't run into issues if, for example, somebody edits content directly on a github repository, which isn't really possible to do with the ipynb files

jstac:
Yes, "markdown takes precedence whenever it exists" is an easy rule for the user to remember.

chrisjsewell:
> A few edge cases:
> 
> - If you remove `name.ipynb` and add `name.md` (i.e. same name, different extension) before running sphinx; this is "confusing" sphinx and breaking the cache (by not removing the `name.ipynb`).
> 
> -  If a text-based notebook fails to execute, and so doesn't get updated with exec metadata, 

Added 'fixes' for these cases

chrisjsewell:
All the documentation notebooks are now markdown files, and I have added a footnote with download links for the files as both markdown and (executed) notebooks: https://318-241268229-gh.circle-artifacts.com/0/html/use/basic.html#download
As discussed in ExecutableBookProject/cli#59, this may be a stop-gap for adding these in a drop-down menu as part of the theme. Plus there is also the whole binder/thebelab links to consider

chrisjsewell:
This can be merged if people are happy

docs/use/basic.md
```diff
@@ -0,0 +1,98 @@
...
+  name: python3
+---
+
+# An example Jupyter Notebook
+
+This notebook is a demonstration of directly-parsing Jupyter Notebooks into
+Sphinx using the MyST parser.[^download]
+
+[^download]: This notebook can be downloaded as
+            **{jupyter-download:notebook}`basic`** and {download}`basic.md`
```
choldgraf, line 20: Any particular reason to use `jupyter-download` here and not just `download`?

docs/use/markdown.md
```diff
@@ -1,27 +1,109 @@
...
+  text_representation:
+    format_name: myst
+kernelspec:
+  display_name: Python 3
+  name: python3
+---
+```
+
+```{tip}
+You can also create the file from an existing notebook: `jupytext notebook.ipynb --to myst`
```
choldgraf, line 38: I think this needs to be `md:myst`

myst_nb/__init__.py
```diff
@@ -66,8 +66,9 @@ def save_glue_cache(app, env):
 
 def setup(app):
     """Initialize Sphinx extension."""
-    # Sllow parsing ipynb files
-    app.add_source_suffix(".ipynb", "ipynb")
+    # Allow parsing ipynb files
+    app.add_source_suffix(".md", "myst-nb")
```
choldgraf, line 70: These are overridden when people manually specify this in `conf.py`, yeah?

myst_nb/converter.py
```diff
@@ -0,0 +1,239 @@
...
+
+    :param text: the file text
+    :param code_directive: the name of the directive to search for containing code cells
+    :param raw_directive: the name of the directive to search for containing raw cells
+    :param add_source_map: add a `source_map` key to the notebook metadata,
+        which is a list of the starting source line number for each cell.
+
+    :raises MystMetadataParsingError if the metadata block is not valid JSON/YAML
+
+    NOTE: we assume here that all of these directives are at the top-level,
```
choldgraf, line 158: This is a pretty big assumption, no? So myst-nb will not support nested directives?

myst_nb/parser.py
```diff
@@ -13,45 +13,56 @@
...
+                "Notebook load failed for %s: %s", self.env.docname, err
+            )
+            return
+        if not ntbk:
+            # Read the notebook as a text-document
+            to_docutils(inputstring, options=self.config, document=document)
+            return
+
         # add outputs to notebook from the cache
         if document.settings.env.config["jupyter_execute_notebooks"] != "off":
```
choldgraf, line 64:
```suggestion
        if self.env.config["jupyter_execute_notebooks"] != "off":
```

it seemed you were doing this elsewhere

docs/use/basic.md
chrisjsewell: Because the file is taken from the `jupyter_execute` folder, i.e. the post-execution notebook, not the source folder as `download` looks at. I guess see `jupyter-sphinx` for the actual logic.

docs/use/markdown.md
chrisjsewell: nope, I added in some sneaky logic into jupytext to intercept `myst` and read it as `md:myst`, so both work.

choldgraf: Hmm - is this in master and not released yet? I just tried it locally on the latest pypi version and it didn't work

myst_nb/__init__.py
chrisjsewell: Yes, as in if you do:

```python
source_suffix = {
   ".ext": "parser_type"
}
```

then now only `.ext` would be read by sphinx. I changed `ipynb` to `myst-nb` because now obviously this covers more than just `.ipynb` extensions

choldgraf: makes sense to me 👍

docs/use/markdown.md
chrisjsewell: Not sure, I using the dev version, so it will definitely work at some point. This conversion won't actually work at the moment with latest myst versions anyway, until mwouts/jupytext#473 is merged and released

docs/use/basic.md
choldgraf: right - I remember now, thanks

myst_nb/parser.py
choldgraf: any reason not to commit this? I'll merge once it's resolved one way or another

myst_nb/converter.py
chrisjsewell: Not all directives, just `{code-cell}`. This is necessary for compliance with round trip conversion with notebooks, as obviously you can't have a code cell nested in a markdown cell. When I say 'assume`, I mean we just ignore any nested code cells, which will, if present, then raise an error on the sphinx parse, because it won't have a directive associated with it.

choldgraf: Ah - that makes sense then, since `{code-cell}` doesn't support myst-markdown

chrisjsewell: Yep thats the one 'compromise' we make, in terms of treating the whole file as a notebook (rather than 'extracting' notebooks from files, like jupyter-sphinx does): you can't do something like

`````
````{note}
```{code-cell}
a = 1
```
````
`````

with the benefit of being abled to treat them exactly the same a standard notebooks; integrating them with jupyter-cache etc.

docs/use/markdown.md
chrisjsewell: Note I don't use jupytext as a dependency at present, as some of the code is a replica of the code in mwouts/jupytext#473, which could be swapped in at some point. To keep it simple, we only allow the myst format for now, not any of the other jupytext formats

choldgraf:
boom! I am very excited, will test out today

PR #121, opened 2020-04-03
adding jupyter notebook mathjax config

choldgraf:
This adds the default notebook mathjax config as a little javascript file.

@phaustin @chrisjsewell think this will fix the problem?

closes https://github.com/ExecutableBookProject/MyST-NB/issues/99

### DIFF
--- a/docs/use/basic.md
+++ b/docs/use/basic.md
@@ -38,7 +38,14 @@ Wow, a note!
 ```
 ````
 `````
-Other MyST features like **equation numbering** can be used in notebooks:
+
+Equations work as expected:
+
+\begin{equation}
+\frac {\partial u}{\partial x} + \frac{\partial v}{\partial y} = - \, \frac{\partial w}{\partial z}
+\end{equation}
+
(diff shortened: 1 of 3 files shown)


phaustin:
Trying to test this, but I've hit a snag at the ipynb stage.  Can someone confirm that 

```
$$
\begin{align}
\frac{\partial v}{\partial t} + u \frac{\partial v}{\partial x} + v \frac{\partial v}{\partial y} + w \frac{\partial v}{\partial z} + fu = - \, \frac{1}{\rho} \, \frac{\partial p}{\partial y}
+ A_v \, \frac{\partial^2 v}{\partial z^2} + A_h \, \nabla^2 v
\end{align}
$$
```
works in jupyter?   The jupyter parser adds a nested equation environment that mathjax chokes on.

choldgraf:
Yep, just tried it and it works OK

phaustin:
ok -- must be the latex notebook extension I've installed in jupyter that's getting in the way

choldgraf:
OK cool - I'm gonna merge this away and we can spot-check bugs if they come up

PR #127, opened 2020-04-04
updating inline glue css

choldgraf:
I was reading [this fivethirtyeight article](https://fivethirtyeight.com/features/why-its-so-freaking-hard-to-make-a-good-covid-19-model/) and thought that they had an interesting approach to using "variables in prose". They set all variables to be mono fonts, and added a colored background to them:

image

This PR takes a step in that direction, adding a slight grey background behind inserted variable values, so that it is more obvious where we have "pasted" variables vs. regular text.

Another possibility would be to make the `paste:text` output an inline node instead of a text node, so it's mono-fonted as well.

An example of how it looks now:

image

from: https://343-241268229-gh.circle-artifacts.com/0/html/use/glue.html

What do folks think?

### DIFF
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -110,8 +110,9 @@ div.cell_output table {
 
 
 /* Inline text from `paste` operation */
-
-span.pasted-text {
+span.pasted-inline code, span.pasted-text {
+  background: #e9ecef;
+  padding: 0 .2em;
   font-weight: bold;
 }
 


chrisjsewell:
I guess ideally this would be easily configurable by the user, is the best way to do this *via* sphinx conf `html_css_files = ['css/custom.css']`, or should/could there be more specific configuration that alters the base `mystnb.css` (i.e. it would be more of a template than a static file)

> Another possibility would be to make the paste:text output an inline node instead of a text node, so it's mono-fonted as well

I don't think this should be done, if you don't want it to be text, then a new role should be added

choldgraf:
I think to start with, we can recommend in the documentation how users can over-ride these specific values to control different behavior. Maybe longer-term we can consider "templatizing" some things, though that feels a bit more complex to me.

For the mono font thing, the main reason I thought of this was because I don't think you can do string formatting with the `:any` role so outputting floats would always have a huge number of decimals. Not something that's a dealbreaker one way or another though.

Do you want me to add in a bit about customizing the CSS to this PR?

choldgraf:
I'm closing this one, it's outdated and anyway I think that special-casing the CSS for pasted text is probably too opinionated as a default option anyway

PR #141, opened 2020-04-07
Add test for ipynb as toctree

chrisjsewell:


chrisjsewell:
@AakashGfude @mmcky In this PR I add a test for including a toctree in a notebook, and don't find any issues?

You need to modify this test to reproduce the issue you are having. 
If you uncomment the following line, with a suitable path on your machine:

https://github.com/ExecutableBookProject/MyST-NB/blob/2cf7e6467eae1c64daf124d4020a29b1b47412da/tests/test_parser.py#L49

Then run `pytest -k test_toctree_in_ipynb`,
it will output the full test build there for you to look at (rather than in the default temporary directory)

chrisjsewell:
The main thing here is that we need a **minimal** working example of the problem, to identify the issue

mmcky:
@chrisjsewell here is a pretty minimal example

https://github.com/ExecutableBookProject/myst-nb.example-project/tree/minimal

mmcky:
So the issue comes about when a notebook contains a `code` block. As soon as I delete the `code cell` from `ar1_processes` document in the above minimal example. The `pdf` will compile as expected without the sphinx warning that the `ar1_processes` document is missing.

chrisjsewell:
From b541b73, this is only an issue for latex (and possibly other non-html) builds.
The initial doctrees are created correctly created, but for some reason `other.ipynb` is not included in the build process in later phases, evidenced by the post-transform `CellOutputsToNodes` never being applied to it.

mmcky:
@chrisjsewell I suspect this will be an issue for:

1. `texinfo`
1. `manpage`
1. `latex`
1. `singlehtml`

not that we plan to support `manpage` and `texinfo`

mmcky:
Ok I think the offending line in `docutils` is:

```python
*** TypeError: __init__() got multiple values for argument 'outputs'
```

when it tries to unpack `**attributes` there are two `outputs` in the method call as `**attributes` includes `{'ids': [], 'classes': [], 'names': [], 'dupnames': [], 'backrefs': [], 'outputs': 0}`

and the method has an `outputs` argument

```
class CellOutputBundleNode(nodes.container):
    """Represent a MimeBundle in the Sphinx AST, to be transformed later."""

    def __init__(self, outputs, rawsource="", *children, **attributes):
        self.outputs = outputs
        attributes["outputs"] = len(outputs)
        super().__init__("", **attributes)
```

chrisjsewell:
Thanks @mmcky, this should be fixed in ebd9363

myst_nb/parser.py
```diff
@@ -230,7 +230,7 @@ def render_nb_code_cell(self, token: Token):
         # ==================
         # Cell output
         # ==================
-        if "remove_output" not in tags:
+        if "remove_output" not in tags and cell["outputs"]:
```
chrisjsewell, line 233: Here I now skip the creation of the `CellOutputNode` if the code cell has no outputs. @choldgraf any reason you can think of why this should not be done?

mmcky:
perfect thanks @chrisjsewell just got the `pdf` building by changing the count variable name.

myst_nb/parser.py
choldgraf: Nah that sounds like a good idea to me

chrisjsewell: 👍

PR #146, opened 2020-04-08
css to align and float figure directives

najuzilu:
I've added some basic css to align figures using `figclass` option.

Here's a quick screenshot of the documentation I'd like to add to the `cli`:
Screen Shot 2020-04-08 at 3 20 55 PM

Would love to hear what people think about this @chrisjsewell @choldgraf

### DIFF
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -117,3 +117,36 @@ span.pasted-text {
 span.pasted-inline img {
   max-height: 2em;
 }
+
+/* Align figures */
+
+div.figure.align-right {
+  display: flex;
+  justify-content: flex-end;
+}
+
+div.figure.align-center {
+  display: flex;
+  justify-content: center;
+}
+
+/* Float figures */
+
+div.figure.float-left {
+  clear: left;
+  float: left;
+  margin-left: 1em;
+  margin-right: 2em;
+}
+
+div.figure.float-right {
+  clear: right;
+  float: right;
+  margin-left: 2em;
+  margin-right: 1em;
+}
+
+.highlight {
+  background: none;
+}
+
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -149,4 +149,3 @@ div.figure.float-right {
 .highlight {
   background: none;
 }
-


chrisjsewell:
Yeh I don't see any problem with that in principle 👍 
You need to add a new line to the bottom of the file to pass pre-commit (see [code style](https://myst-nb.readthedocs.io/en/latest/develop/contributing.html#code-style) for details)

Also why you are at it, it would be good to add some documentation on this option, and also on the CSS in general, e.g. a note on how people can override it: https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html

choldgraf:
this sounds great to me - the one thing I'd double-check is that the aligning CSS works across multiple github themes (and in particular for the sphinx-book-theme)

najuzilu:
@chrisjsewell should I include this information in the `MyST-NB` doc or `cli` doc?
@choldgraf it works great with `sphinx-book-theme`. I'll double check that it works across multiple github themes too.

choldgraf:
I think we'll have to do both, for the documentation

chrisjsewell:
Yep add it here first and we’ll take it from there

choldgraf:
@najuzilu I just realized that this fix is probably better in the [`pydata-sphinx-theme`](https://github.com/pandas-dev/pydata-sphinx-theme) as well as the `sphinx-book-theme`, rather than being in MyST-NB, because these fixes are not unique to Jupyter Notebooks or their outputs.

I think the path forward would be to:

1. From the `sphinx-book-theme` take the "standard" figure CSS rules (the `align-` ones that are triggered with `:align:` and the "float" ones that are triggered with `:figclass:` and *move* them into `pydata-sphinx-theme` (since the book theme inherits from pydata, we'll get the CSS from there).
2. For the "extra" CSS classes (AKA, the sidebar ones), put those in the `sphinx-book-theme`. They feel a bit more customized to the book theme to me.

najuzilu:
@choldgraf sounds good. Will do this now. :-)

choldgraf:
hey all - if I recall this is now in sphinx-book-theme, is that right @najuzilu ?

najuzilu:
@choldgraf yes. I'll go ahead and close this issue.

PR #147, opened 2020-04-08
serializing nested ntbk metadata items

choldgraf:
I *think* that this is the only thing that we need to do in order to get nested metadata working.

edit: I think I need to edit some of the tests because the metadata in the output document will be different...I will try and figure out to update the regressions stuff so that it works (though somebody comment if that's not the problem here). @chrisjsewell are we OK with just updating the regression templates so that they now include the JSON-serialized nested metadata?

closes #147

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -6,7 +6,16 @@ def test_basic_run(sphinx_run, file_regression):
     sphinx_run.build()
     # print(sphinx_run.status())
     assert sphinx_run.warnings() == ""
-    assert sphinx_run.app.env.metadata == {"basic_run": {"test_name": "notebook1"}}
+    assert set(sphinx_run.app.env.metadata["basic_run"].keys()) == {
+        "test_name",
+        "kernelspec",
+        "language_info",
+    }
+    assert sphinx_run.app.env.metadata["basic_run"]["test_name"] == "notebook1"
+    assert (
+        sphinx_run.app.env.metadata["basic_run"]["kernelspec"]
+        == '{"display_name": "Python 3", "language": "python", "name": "python3"}'
+    )
     file_regression.check(sphinx_run.get_doctree().pformat(), extension=".xml")
 
     filenames = {
@@ -20,11 +29,28 @@ def test_basic_run(sphinx_run, file_regression):
 )
 def test_complex_outputs(sphinx_run, file_regression):
     sphinx_run.build()
-    # print(sphinx_run.status())
     assert sphinx_run.warnings() == ""
-    assert sphinx_run.app.env.metadata == {
-        "complex_outputs": {"celltoolbar": "Edit Metadata", "hide_input": "False"}
+
+    assert set(sphinx_run.app.env.metadata["complex_outputs"].keys()) == {
+        "ipub",
+        "hide_input",
+        "nav_menu",
+        "celltoolbar",
+        "latex_envs",
+        "kernelspec",
+        "language_info",
+        "jupytext",
+        "toc",
+        "varInspector",
     }
+    assert (
+        sphinx_run.app.env.metadata["complex_outputs"]["celltoolbar"] == "Edit Metadata"
+    )
+    assert sphinx_run.app.env.metadata["complex_outputs"]["hide_input"] == "False"
+    assert (
+        sphinx_run.app.env.metadata["complex_outputs"]["kernelspec"]
+        == '{"display_name": "Python 3", "language": "python", "name": "python3"}'
+    )
     file_regression.check(sphinx_run.get_doctree().pformat(), extension=".xml")
 
     filenames = {
--- a/myst_nb/parser.py
+++ b/myst_nb/parser.py
@@ -117,7 +117,7 @@ def nb_to_tokens(ntbk: nbf.NotebookNode) -> Tuple[MarkdownIt, AttrDict, List[Tok
 
     for cell_index, nb_cell in enumerate(ntbk.cells):
 
-        # if the the source_map ahs been stored (for text-based notebooks),
+        # if the the source_map has been stored (for text-based notebooks),
         # we use that do define the starting line for each cell
         # otherwise, we set a pseudo base that represents the cell index
         start_line = source_map[cell_index] if source_map else (cell_index + 1) * 10000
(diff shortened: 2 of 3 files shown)


chrisjsewell:
Yep the failed tests show that the metadata is being correctly propagated to the sphinx env, you just need to update them!

choldgraf:
OK I think I've got it fixed (in a fairly inelegant way, but 🤷‍♂ ). I had to update the pre-commit flake8 test because the nested JSON lines are too long, but other than that tests were passing

tests/test_parser.py
```diff
@@ -23,7 +29,18 @@ def test_complex_outputs(sphinx_run, file_regression):
...
+        "complex_outputs": {
+            "celltoolbar": "Edit Metadata",
+            "hide_input": "False",
+            "ipub": '{"bibliography": "example.bib", "biboptions": ["super", "sort"], "bibstyle": "unsrtnat", "language": "portuges", "listcode": true, "listfigures": true, "listtables": true, "titlepage": {"author": "Authors Name", "email": "authors@email.com", "institution": ["Institution1", "Institution2"], "logo": "logo_example.png", "subtitle": "Sub-Title", "supervisors": ["First Supervisor", "Second Supervisor"], "tagline": "A tagline for the report.", "title": "Main-Title"}, "toc": {"depth": 2}, "pandoc": {"use_numref": true, "at_notation": true}, "sphinx": {"bib_title": "My Bibliography"}}',
+            "jupytext": '{"metadata_filter": {"notebook": "ipub"}}',
+            "kernelspec": '{"display_name": "Python 3", "language": "python", "name": "python3"}',
+            "language_info": '{"codemirror_mode": {"name": "ipython", "version": 3}, "file_extension": ".py", "mimetype": "text/x-python", "name": "python", "nbconvert_exporter": "python", "pygments_lexer": "ipython3", "version": "3.6.2"}',
```
choldgraf, line 38: @chrisjsewell is there any reason for us to think any of this stuff will be brittle and break over time? (e.g. as libraries update etc?)

choldgraf: ugh I think the answer is yes, because the tests are erroring because the version of jupytext they've got is different from the one on my machine. hmmmmm

.pre-commit-config.yaml
```diff
@@ -6,7 +6,9 @@ exclude: >
       \.vscode/settings\.json|
       tests/commonmark/commonmark\.json|
       .*\.xml|
-      tests/.*\.txt
+      tests/.*\.txt|
+      tests/test_parser.py|
+      tests/test_text_based.py
```
chrisjsewell, line 11: Lazy 😜 ! Add `# noqa: E501` to specific lines that are too long, not ignore the whole file

tests/test_parser.py
```diff
@@ -23,7 +29,18 @@ def test_complex_outputs(sphinx_run, file_regression):
     # print(sphinx_run.status())
     assert sphinx_run.warnings() == ""
     assert sphinx_run.app.env.metadata == {
```
chrisjsewell, line 31: as this is pretty long now, probably easier to add a `data_regression.check`

chrisjsewell: same with the others

.pre-commit-config.yaml
choldgraf: ahhh that's right, I knew there was a way to do that

tests/test_parser.py
choldgraf: ok this is a good opportunity to learn about `data_regression.check`. I guess that we will just have to update the regression templates any time that the jupytext version is updated

chrisjsewell: Yeh, or if it's going to change to often, just do as a minimum `assert set(sphinx_run.app.env.metadata.keys()) == {"celltoolbar", ...}`

choldgraf:
@chrisjsewell where does the jupytext dependency get defined? I didn't see it in there, but I think it's needed for myst text reading, no? Are we inheriting it from some other package? And if so then should I just add it in explicitly?

chrisjsewell:
> where does the jupytext dependency get defined? I didn't see it in there, but I think it's needed for myst text reading, no? Are we inheriting it from some other package? And if so then should I just add it in explicitly?

There's (at least currently) no specific dependency of `myst-nb` or jupytext, since the [relevant code from jupytext](https://github.com/mwouts/jupytext/blob/b8693695640efa0659d91ab6f5a2524ca9f75471/jupytext/myst.py#L213) is essentially copied into `myst_nb/convert`. It's an open question if, now that jupytext 1.4.2 is released, we delete this code and import from jupytext.

The versions you see in the notebook/myst-markdown metadata, are/were generated during `jupytext --to myst notebook.ipynb` or when you pair the notebook in jupyter notebook/lab.
They don't change dynamically and are not strictly required for reading in the current code (jupytext uses them to checks for compatible versions: https://github.com/mwouts/jupytext/blob/b8693695640efa0659d91ab6f5a2524ca9f75471/jupytext/formats.py#L365). So yeh they are static and won't change if dependencies change

choldgraf:
OK take a look at the latest testing suite. I'm checking for all the keys in the metadata, as well as for the proper values for a few keys, including both "flat" and "nested" values. Think that works?

tests/test_parser.py
```diff
@@ -6,7 +6,16 @@ def test_basic_run(sphinx_run, file_regression):
...
-    assert sphinx_run.app.env.metadata == {"basic_run": {"test_name": "notebook1"}}
+    assert set(sphinx_run.app.env.metadata["basic_run"].keys()) == {
+        "test_name",
+        "kernelspec",
+        "language_info",
+    }
+    assert sphinx_run.app.env.metadata["basic_run"]["test_name"] == "notebook1"
+    assert (
+        sphinx_run.app.env.metadata["basic_run"]["kernelspec"]
+        == '{"display_name": "Python 3", "language": "python", "name": "python3"}'
```
choldgraf, line 17: note - black did this, and it seemed reasonable + removed the line too long issues, so I went with it

choldgraf:
image

chrisjsewell:
Yep looks fine to me 👍

chrisjsewell:
I don't think you realize how easy you have it here lol. For example, I just has this PR merged https://github.com/aiidateam/aiida-core/pull/3613 after months of having to address revisions from 3 reviewers and a lot more stringent CI checks 😬

choldgraf:
haha - my worst was in switching over the matplotlib documentation entirely to use sphinx-gallery. It took 7 months and more bikeshedding than I have ever experienced in my life :-)

PR #150, opened 2020-04-13
github action conditions

choldgraf:
I found that we have duplicate tests running because our github action runs on both pull_request and push. I believe that this configuration will make it run only once for forked pull requests so we don't duplicate test jobs

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,12 @@
 name: continuous-integration
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
 
 jobs:
   pre-commit:


choldgraf:
I'm not gonna merge this just in case there was a specific reason that @chrisjsewell configured it the way he did before

chrisjsewell:
Yeh it's a *known* issue. It's not an issue with "forked" PRs, as in from branches on `OtherAccount/MyST-NB`,
just with internal PRs, as in from other branches on `ExecutableBookProject/MyST-NB`. Then you trigger for the commit to that branch and for the merge into `main`.

The problem with your configuration is that, if you are working on another branch, you now won't get CI feedback until you create the PR to `main`. This isn't ideal in the use case, for example, where you are working on a particularly big change that you don't want to make a PR for straight away, and you can't do things like make a PR to a non-main branch.

Ideally you would conditionally trigger for `push` only if `pull-request` is not triggered. I looked at this before, but it doesn't seem possible:

- https://github.community/t5/GitHub-Actions/How-to-trigger-an-action-on-push-or-pull-request-but-not-both/td-p/35805
- https://github.community/t5/GitHub-Actions/How-to-prevent-running-actions-on-push-when-pull-request-is-also/td-p/47013

So in summary, fine for you to merge this, but I reserve the right to change it back if necessary later down the line lol

choldgraf:
sounds good - and that makes sense. I don't feel *super* strongly about it, just seems like we waste a little bit of time running twice the jobs. I'm happy to move it back if it would be useful when the time comes 👍

PR #153, opened 2020-04-17
Added a copy method to celloutputbundlenode

AakashGfude:
Added a copy method which also takes `outputs` as argument

fixes https://github.com/ExecutableBookProject/MyST-NB/issues/151

### DIFF
--- a/myst_nb/parser.py
+++ b/myst_nb/parser.py
@@ -259,6 +259,11 @@ class CellOutputBundleNode(nodes.container):
         attributes["output_count"] = len(outputs)
         super().__init__("", **attributes)
 
+    def copy(self):
+        return self.__class__(
+            rawsource=self.rawsource, outputs=self.outputs, **self.attributes
+        )
+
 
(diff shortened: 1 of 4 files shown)


choldgraf:
Is this something that we could add a test for?

AakashGfude:
@choldgraf we can test this as a part of a `latex` build in `test_transform`?

choldgraf:
sure - I just wanna make sure that the bug we think we are fixing doesn't re-surface :-)

AakashGfude:
Hey @choldgraf , what do you think about the test added?

choldgraf:
it looks good to me - is it a test that would have failed w/o the `copy` method?

AakashGfude:
@choldgraf , yes it would not have been able to replace `<CelloutputBundleNode>` with specific mime-type nodes.

choldgraf:
Great - I'll merge tomorrow AM unless somebody else suggests otherwise 👍 nice job @AakashGfude !

Issue #126, opened 2020-04-03
Ensure Math Is Parsed/Rendered Consistently


phaustin:
To anchor our choices in some concrete examples, I've pushed #125 
which creates https://phaustin.github.io/MyST-NB/equno.html from
https://github.com/phaustin/MyST-NB/blob/eqno/docs/examples/eqno_test/equno.md

Questions:  

1) do we want or need similar behviour between latex-rendered and html-rendered versions of the same math?
2) What's the best way to implement the behavior we expect from 1)

(prelimnary aside -- @choldgraf -- as the html shows, I'm not getting mystnb.js to actually run on the page and toggle processEnvironments=true -- can you confirm that using the myst environment from eqno_test/environment.yml does properly fomat the no-$$ equation?  I see msytnb.js in _build/_static but it doesn't appear to do anything.)

phaustin:
@chrisjsewell to check my understanding -- when you say "write a plugin" would it be a python port of  this markdown-it mathjax extension?  https://www.npmjs.com/package/markdown-it-mathjax

chrisjsewell:
Yep something like that 👍 , we only need a new plugin that handles the latex environments component though, not the dollars/brackets are already handled by the https://github.com/goessner/markdown-it-texmath port, see https://github.com/ExecutableBookProject/markdown-it-py/tree/master/markdown_it/extensions/texmath

phaustin:
that would be great, have no strong opinion on katex vs. mathjax 3 -- will note that some of our math-heavy mathjax 2.7 jupyer notebooks take a full 30 seconds to render with a 2019 macbook pro, so moving on from 2.7 was something we were looking at.

choldgraf:
Just a note that I had a conversation about katex vs. mathjax with @rowanc1 and he mentioned that katex was way (way, way, way) faster for HTML rendering

phaustin:
yes, although mathjax 3 claims it's doubled it's rendering speed, I gues that's not way**4 though

chrisjsewell:
Yeh I also have got the impression that katex is the way forward these days

rowanc1:
Interesting, that advice was pre-mathjax 3, which looks a lot better!

There are some detailed comparisons here:
https://www.intmath.com/cg5/katex-mathjax-comparison.php

Timings for me were: 314 vs 1341 vs 460 (katex, mathjax 2, mathjax 3)

There is a flicker on the mathjax loading which isn't there on katex, I think it is the font loading.

[...]

PR #159, opened 2020-04-22
myst parser stripping one backslash from bare align equations

phaustin:
The new mystnb.js mathjax config is working correctly, and myst-generated html for align environments can be displayed without $$ delimters now that processEnvironments is set to true.  There is one problem remaining -- the myst parser treats `\\` in bare environments as a single escaped `\\` -- specifically [these lines](https://github.com/phaustin/MyST-NB/blob/3045d54faf0ae9cfd9831412d2d7aecf7b1716cc/docs/examples/eqno_test/equno_myst.html#L135-L139) which produce [this page](https://phaustin.github.io/MyST-NB/equno.html).  Hand editing to replace the stripped `\` produces a page that displays correctly.

To reproduce:  run [build_website.sh](https://github.com/phaustin/MyST-NB/blob/align_eq/docs/examples/eqno_test/build_website.sh)


chrisjsewell:
and this is perfectly "correct" behaviour from the markdown parser ([see here](https://spec.commonmark.org/0.15/#backslash-escapes)), which is why this parsing of the "bare" math environments needs to be handled at the markdown parser level, and not by mathjax, as discussed in https://github.com/ExecutableBookProject/cli/pull/73#issuecomment-616020781 and #126

chrisjsewell:
BTW its best to open these "example" PR as drafts 😁 https://github.blog/2019-02-14-introducing-draft-pull-requests/

phaustin:
thanks, that's a useful new github feature.  Also, I now see from the [the mathjax docs](http://docs.mathjax.org/en/latest/options/input/tex.html) that  processEnvironments is true by default, the mathjax config was a red herring -- this was the actual source of the difference between jupyter/nbconvert/nbsphinx and markdown/myst.

PR #165, opened 2020-04-23
Checking image paths for each cell in transform(needed for latex)

AakashGfude:
**Note :-  The function `cell_output_to_nodes` is copied from https://github.com/jupyter/jupyter-sphinx for demonstration purposes only. And this PR is not intended to be merged.**

Hey @choldgraf @mmcky , I have added a few harmless lines of code to the `cell_output_to_nodes`  function (also added to `cell_output_to_nodes_inline`), to target the `image file not readable` error in https://github.com/ExecutableBookProject/cli/issues/115 . It is beneficial for `latex` builds in scenarios where the source directory has subdirectories. Won't affect anything for `html` or other builds.
If it looks ideal, maybe we can create a PR for this at https://github.com/jupyter/jupyter-sphinx ?


myst_nb/transform.py
```diff
@@ -63,6 +62,136 @@ def apply(self):
...
+                continue
+            data = output["data"][mime_type]
+            if mime_type.startswith("image"):
+                # Sphinx treats absolute paths as being rooted at the source
+                # directory, so make a relative path, which Sphinx treats
+                # as being relative to the current working directory.
+                filename = os.path.basename(output.metadata["filenames"][mime_type])
+
+                # checks if path in cell has a subpath inside of dir
+                filedir = os.path.dirname(output.metadata["filenames"][mime_type])
```
AakashGfude, line 148: these five lines

choldgraf: It seems reasonable to me @akhmerov what do you think?

akhmerov: Seems reasonable to me as well, except I don't quite understand the magical `filedir.split(dir)`. @AakashGfude can you elaborate more? Under which conditions is `len(subpaths) > 1`? Is `subpaths` ever empty?

AakashGfude: @akhmerov sure.  So, the below code will help in explaining 
image


As you can see, if `subpaths` exist, they are never empty. They have a size greater than one if the path passed as argument is a subpath.

AakashGfude:
@choldgraf @mmcky I have modified this PR and removed the local declaration of `cell_output_to_nodes`. Changes to `cell_output_to_nodes_inline` are required, so have kept that.

This branch should be good to **merge** now.

choldgraf:
@AakashGfude jupyter-sphinx just released a new version, which I gather is related to this PR? If so then you should feel free to merge this PR, but I won't do so in case you think there is something else to be done

AakashGfude:
@choldgraf  nicee. This PR is ready to be merged from my side. Thanks

AakashGfude:
fixes https://github.com/executablebooks/jupyter-book/issues/505

PR #167, opened 2020-05-02
Added a simple {tableofcontents} directive

AakashGfude:
**NOTE** : This branch is work in progress. The implementations can change drastically.

@choldgraf @jstac @mmcky @chrisjsewell @najuzilu 

@choldgraf I know you mentioned here https://github.com/executablebooks/jupyter-book/issues/503, that we don't necessarily need a `{toctree}` node, but I wanted to try out a basic implementation of it, just for the ease with which it creates titles and links under the hood.

Do you think it can create complications in the future?

It has neat features though , which we can leverage.  For example :- 

1. 

Code :- 

```
```{tableofcontents}
```

Result :- 

image

2. 

Code :- 

```
```{tableofcontents}
:maxdepth: 2
```

Result :- 

image

### DIFF
--- a/myst_nb/__init__.py
+++ b/myst_nb/__init__.py
@@ -19,6 +19,7 @@ from .parser import (
     CellOutputNode,
     CellOutputBundleNode,
 )
+from .directives import toc
 from .transform import CellOutputsToNodes
 from .nb_glue import glue  # noqa: F401
 from .nb_glue.domain import NbGlueDomain
@@ -120,6 +121,15 @@ def setup(app):
         man=(skip, None),
     )
 
+    app.add_node(
+        toc.TableOfContentsNode,
+        override=True,
+        html=(toc.visit_TableOfContentsNode, toc.depart_TableOfContentsNode),
+        latex=(toc.visit_TableOfContentsNode, toc.depart_TableOfContentsNode),
+        textinfo=(toc.visit_TableOfContentsNode, toc.depart_TableOfContentsNode),
+        text=(toc.visit_TableOfContentsNode, toc.depart_TableOfContentsNode),
+        man=(toc.visit_TableOfContentsNode, toc.depart_TableOfContentsNode),
+    )
     # Add configuration for the cache
     app.add_config_value("jupyter_cache", "", "env")
     app.add_config_value("execution_excludepatterns", [], "env")
@@ -139,6 +149,7 @@ def setup(app):
     app.add_js_file("mystnb.js")
     app.setup_extension("jupyter_sphinx")
     app.add_domain(NbGlueDomain)
+    app.add_directive("tableofcontents", toc.TableofContents)
 
     # TODO need to deal with key clashes in NbGlueDomain.merge_domaindata
     # before this is parallel_read_safe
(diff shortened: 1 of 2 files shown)


choldgraf:
Nice! As long as it doesn't mess with the builds then I think it's fine. I just wondered whether there are post-processors etc that look for TOCtree objects in pages

AakashGfude:
> I just wondered whether there are post-processors etc that look for TOCtree objects in pages
I will keep an eye out for that

AakashGfude:
Have scrapped the use of `{toctree}` node here, because of complications, and instead have used basic docutils and sphinx nodes. 

At present, only catered for `jupyter-book` project with `_toc.yml` file. 

@choldgraf  Since we are going to document this directive in `myst_nb`, should we handle for cases where people declare this directive in their sphinx projects where they use `myst_nb` but not have `_toc.yml`, by raising an error?

Using this directive in `jupyter-book` documentation page - https://jupyterbook.org/intro.html , gives the following view :- 

image

Owing to the design of the variable `self.config.globaltoc`, which stores the entries of `_toc.yml`, this toc has `Books with Jupyter` (intro.html) as the main parent.
To prevent this. should I always remove the `master_doc` from the toc?

Is their a page stating the conventions `_toc.yml` should follow? 


Styling is not done yet. Will implement it in mystnb.css.

cc: @mmcky @jstac

choldgraf:
Looking good! A couple quick thoughts:

* Since this will only make sense for use in `jupyter-book`, why not add this as a PR to jupyter-book rather than myst-nb?
* The way I'd imagine this being used is: if used on a page, grab the `TocTree` for the page and display all second-level-and-below options (so the current-page item isn't shown). Maybe (and doesn't need to be in this PR) allow for a relative path to another page as an argument, and if so then show the toctree for *that* page instead.
* We could style things so that page captions don't get a bullet-point and instead look more like "headers" with pages underneath them.

AakashGfude:
> * Since this will only make sense for use in `jupyter-book`, why not add this as a PR to jupyter-book rather than myst-nb?

You are right. Now that we have established this, it makes more sense to put it there. 

> * The way I'd imagine this being used is: if used on a page, grab the `TocTree` for the page and display all second-level-and-below options (so the current-page item isn't shown). Maybe (and doesn't need to be in this PR) allow for a relative path to another page as an argument, and if so then show the toctree for _that_ page instead.

So we won't have a static toc directive but a dynamic one depending on the page, and only showing its children links? We can also keep this programmable? like `full` for the whole toc and `default` is like the one you mentioned? I am just throwing ideas here, feel free to scrap these. :grin: 


>   * As a side note - if we expand the functionality of `{tableofcontents}` to be more than just "`toctree` but without altering the TOC structure", then I _could_ see it being useful outside the context of jupyter-book...and in that case maybe it could be in its own repository rather than in myst-nb? (e.g. `sphinx-tableofcontents`)

That would be great. 


> * We could style things so that page captions don't get a bullet-point and instead look more like "headers" with pages underneath them.

We can do that. Given that "headers" are always meant to be at the top level, and not in nested structures right?

PR #168, opened 2020-05-04
reorganizing and updating docs

choldgraf:


choldgraf:
tests were happy just before that last push, github actions is degraded right now which is why the tests are hanging, so gonna merge this since it's only a docs update

PR #173, opened 2020-05-09
utf8 encoding issues on windows

phaustin:
[this notebook](https://github.com/phaustin/MyST-NB/blob/utf_issue/docs/examples/utf_bug/notebooks/apdx1_install.md) has the macos cloverleaf character (\u2318).  Running sphinx-build fails [with this error](https://github.com/phaustin/MyST-NB/blob/utf_issue/docs/examples/utf_bug/windows_log.txt).  If I force jupyter-sphinx to use utf encoding with https://github.com/jupyter/jupyter-sphinx/pull/125 I get a successful build.


seanpue:
Hi @phaustin What are your windows and python versions? I thought it was supposed to utf-8 by default in python 3, but maybe that's just for source code.

phaustin:
It's a windows 10 enterprise insider build (slow ring): version 2004
python 3.7.6 from conda-forge
we've had the same problem with another machine that's running the stock release.

phaustin:
just checked the region settings for the current locale (English (United States)).  There's a checkbox that says:
"Beta: Use Unicode UTF-8 for worldwide language support"
but it's not checked -- so I guess UTF-8 is planned but is not the default yet.  (Note that I'm not a windows user, I just keep a PC at home so I can test student labs).

seanpue:
Oh okay. I think that adding the UTF-8 encoding is fine, provided it is added in all file writes. It's the expected encoding, and should anyone need to convert it to UTF-16 or anything, they can add the conversion to their pipeline.

chrisjsewell:
FYI the default encoding is whatever https://docs.python.org/3/library/locale.html#locale.getpreferredencoding returns

seanpue:
Oh cool, thanks @chrisjsewell. @phaustin Could you try it then with PYTHONIOENCODING=utf8 environmental variable set?  That may just affect stdin and stdout, though.

seanpue:
Actually there are some other ones as, including PYTHONUTF8=1: https://docs.python.org/3/using/cmdline.html

phaustin:
i.e. with a powershell script before calling sphinx-build?

seanpue:
Sorry, my windows fu is not what it used to be. You can set it before running sphinx from the command line. There is probably a way to set that environmental variable in your sphinx’s conf.py, as well.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

phaustin:
That works -- here's the successful script:
https://github.com/phaustin/MyST-NB/blob/utf_issue/docs/examples/utf_bug/build_website.ps1

PR #193, opened 2020-06-08
fixing ipywidgets

### LINKED ISSUE #192: Get ipywidgets working
Now that the nbclient widgets bug has been fixed (I think), I think MyST-NB should work with widgets, but it seems that this is not the case. We currently use the `jupyter-sphinx` widgets code to insert widgets into the doctree. This is done:

* by checking if widgets are in a notebook and, if so, inserting a `jupyter_widget_state` token here: https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/parser.py#L172
* That token is then rendered into a `jupyter_sphinx` `JupyterWidgetState` node here: https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/parser.py#L193
  - (which seems to be the same way that `jupyter-sphinx` uses it here: https://github.com/jupyter/jupyter-sphinx/blob/3c9140939db9d683b1f036a55ae4041a6b06d195/jupyter_sphinx/execute.py#L223)
* Note that this should only work with the pre-release version of `nbclient`

When I run MyST-NB, it correctly inserts the `jupyter_widget_state` token, so it seem that the `ipynb` file *does* have widget state inside it. But, the widgets still don't show up for me, and I'm not sure why. Trying to add some context above in case others have an idea for how to debug. Maybe @akhmerov has an intuition?

choldgraf:
This fixes ipywidgets by putting the widgetstatenode inside the doctree rather than prepending to the very top of it (which was causing widgets to be missed because they came before `docinfo`). It also adds a basic test for widgets.

note: I'm adding a new notebook to our test suite (`interactive.ipynb`) because ipywidgets randomly generates widget model IDs and this was causing our regression tests to fail. I think it may also be worth having a notebook specifically for interactive outputs as these can be more complex than traditional outputs.

You can see the widgets working here: https://506-241268229-gh.circle-artifacts.com/0/html/examples/interactive.html#ipywidgets

closes #192 
closes https://github.com/executablebooks/jupyter-cache/issues/31

### DIFF
--- a/docs/examples/interactive.md
+++ b/docs/examples/interactive.md
@@ -104,6 +104,13 @@ You may also run code for Jupyter Widgets in your document, and the interactive 
 outputs will embed themselves in your side. See [the ipywidgets documentation](https://ipywidgets.readthedocs.io/en/latest/user_install.html)
 for how to get set up in your own environment.
 
+```{admonition} Widgets often need a kernel
+Note that `ipywidgets` tend to behave differently from other interactive viz libraries. They
+interact both with Javascript, and with Python. Some functionality in `ipywidgets` may not
+work in default Jupyter Book pages (because no Python kernel is running). You may be able to
+get around this with [tools for remote kernels, like thebelab](https://thebelab.readthedocs.org).
+```
+
 Here are some simple widget elements rendered below.
 
 ```{code-cell} ipython3
@@ -131,6 +138,5 @@ tab.titles = [str(i) for i in range(len(children))]
 tab
 ```
 
-```{code-cell} ipython3
-
-```
+You can find [a list of possible Jupyter Widgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html)
+in the jupyter-widgets documentation.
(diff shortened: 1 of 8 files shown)


choldgraf:
ok the latest push moves the tests back in to `complex_outputs.ipynb` because I found a way to generate ipywidgets that are deterministic (by using a very minimal widget). I think this is ready to go if tests are happy

ofajardo:
Just as an observation, I don't see the tab titles in your example: 

https://506-241268229-gh.circle-artifacts.com/0/html/examples/interactive.html#ipywidgets

choldgraf:
ah nice - I think you're right, I didn't have the code right. Latest push should fix it and if it looks good then I'll merge away

choldgraf:
cool, I think it works!

chrisjsewell:
nice ta, @choldgraf it looks like the only example that not yet working is plotly

choldgraf:
damnit! how did that happen?!

minhhoang1023:
@choldgraf Are you able to render ipywidgets in HTML? When I download it as HTML, it's show as this, and the layout is not shown...

Issue #36, opened 2020-02-24
Deal with markdown outputs of cells


choldgraf:
We don't currently deal with the case that cells output markdown content, we should do so!

Would we want the markdown to become part of the top-level doctree, or would we want it to remain inside the `CellOutputNode` container it is a part of?

originally mentioned in #30

akhmerov:
I think taking these out of the `CellOutputNode` would make the doctree too confusing.

If the appearance of the cells and outputs is easy to control (e.g. hide any extra boxes or output markers), I hope there wouldn't be a problem.

choldgraf:
@akhmerov well I think the problem would be things like we had originally ran into in this repository. If the markdown parser finds a heading, then in the page `doctree` it will create a new section at the top level and start appending to that. So if someone output markdown that had headings in it, then the parser would normally start pulling that out of the `CellOutputNode` it's supposed to be in.

Maybe if we made the assumption: any document structure that is *generated* from a code cell should *not* be a part of the top-level doctree, then we could render that markdown into docutils nodes at transform time using a new `Document` object, and then just do `cell_output_node.children = document.children`

chrisjsewell:
closed with https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#markdown

PR #198, opened 2020-06-14
A 🩹 to include text/markdown outputs in a document

tonyfast:
Currently, text/markdown outputs are skipped, they should not be. Commonly, we generate markdown reports and it would be nice to generate myst markdown.

This pull request adds the ability for text/markdown output to play in sphinx/myst party. Trivially, adding text/markdown to the sphinx transform can in fact render markdown, but it loses scope of the sphinx application. 

This pull requests is a band-aid for the text/markdown output to join to party. The best solution I could come up with was splitting cells when a text/markdown mimetype was discovered. It builds the markdown_it tokens as it moves through the cell outputs.

BTWs, cool project I can't wait to replace recommonmark and nbsphinx all at once. 💟


chrisjsewell:
Hey @tonyfast thanks for the PR. This is certainly something we want to support (see https://github.com/executablebooks/MyST-NB/issues/36), however...
The one main issue I see in the code you have provided, is that it invokes builder specific logic at the parse stage. With the code thus far, we keep the document AST "builder agnostic", until post-transform time.
The reason for this is that the AST is processed and cached once, up to a point just before the post-transforms, so it can be re-used by all builders. Then the post-tranforms are applied per builder (without caching), i.e.
`parse -> apply transforms -> cache -> apply post-transforms -> render with builder`
Therefore, IMO, this is a good start, but I'm not sure if it's the ideal answer yet.

tonyfast:
Thanks for the feedback @chrisjsewell . I made some updates to the PR and added a test.

I don't know if the markdown can be deferred until the renderer when using toctree, I don't think sphinx will ever see.

Y'alls efforts in markdown with myst and markdown_it are very strong. I appreciate the efforts. I'd like to really take full advantage of myst in sphinx by generating documents and reports. In this PR, I have pushed the opinion that Markdown displays are a first class citizens in the notebook alongside the markdown cell.

chrisjsewell:
> I have pushed the opinion that Markdown displays are a first class citizens in the notebook alongside the markdown cell.

Indeed most likely markdown output will have to be a special case. Its just a bit fiddly how to handle it, and definitely needs some thought to ensure all/most edge cases are handled.

> I don't know if the markdown can be deferred until the renderer when using toctree

Oh I see in the test you want to add toctree directives in the markdown output, thats interesting.
Perhaps it would be good if you could provide/link to some of your full notebooks where you use this, to get a better idea of what you would like to achieve.

also cc @choldgraf and @akhmerov for comment

choldgraf:
I think this is a cool addition! One use-case that I can think of is in generating things like "team docs" pages. E.g. here: https://executablebooks.org/en/latest/team.html we have in `conf.py` some code that pulls the latest members of a GitHub team and generates a `.txt` file with MyST markdown. This file is then inserted into the docs with `.. include::` (well, ` ```{include}` ` ;-)).

If we could support markdown outputs and have them parsed at the same time as the rest of the markdown cells, then we could convert that workflow to being a notebook cell that generates the markdown in the output just below, and we could then have that logic on the page, rather than in `conf.py`. Seems cleaner that way.

And also, I could imagine people doing some other cool stuff in that vein if they could programmatically generate MyST markdown (maybe we should call this feature `markmarkdowndown` lol).

tonyfast:
I love some of those use cases you highlight @choldgraf they are spot on.

 I am trying to use myst with [`pidgy`](https://github.com/deathbeds/pidgy), a literate computing shell and kernel for notebooks and markdown files. `pidgy` tangles code from markdown to python ([markdownit_py is used to do this.](https://nbviewer.jupyter.org/github/deathbeds/pidgy/blob/master/pidgy/tangle.ipynb)), and weave the markdown into a display generated with jinja templates. In a pidgy document, every cell is a markdown cell, code cells are on and markdown cells are off. A pidgy notebook will have remove-input on every cell, and only the output is share; it decouples the input from the output. 

`pidgy` is focused on a generating output so I can imagine doing crazy things with myst.

Any of y'all have an ways I could solve this failing test. Do I need to generated a new xml baseline file?

choldgraf:
ah yes - we are using pytest-regression to check that the docutils outputs remain the same. When we *expect* those outputs to change, the way to make the tests pass is:

1. Delete the `.xml` file that we expect to change. (in this case, I think it's `test_glue/test_parser.xml`)
2. Re-run the tests. They will fail saying that the expected file doesn't exist, and the file will be auto-created
3. Re-run the tests, and they will now pass.
4. Commit the new `xml` file to the repository along with your changes

tonyfast:
Thanks for the tip @choldgraf it looks like that did the trick.

myst_nb/parser.py
```diff
@@ -112,6 +113,33 @@ def parse_block(src, start_line):
...
+        tokens = [
+            Token(
+                "nb_code_cell",
+                "",
+                0,
+                meta={"cell": cell},
+                map=[start_line, start_line],
+            )
+        ]
+        for i, output in enumerate(cell["outputs"]):
```
choldgraf, line 126: could you add some comments for what this section is doing (e.g. "If outputs contain markdown, turn it into a new markdown cell to be parsed")

myst_nb/parser.py
```diff
@@ -112,6 +113,33 @@ def parse_block(src, start_line):
...
+            if output["output_type"] == "display_data":
+                if "text/markdown" in output["data"]:
+                    new_code_cell = deepcopy(cell)
+                    new_code_cell["metadata"]["tags"] = new_code_cell["metadata"].get(
+                        "tags", []
+                    ) + ["remove-input"]
```
choldgraf, line 132: Wouldn't we want to create a new *markdown* cell instead of a new code cell? Then not worry about outputs etc because there wouldn't be any for a markdown cell

myst_nb/transform.py
```diff
@@ -195,3 +197,19 @@ def cell_output_to_nodes_inline(
                 to_add.append(JupyterWidgetViewNode(view_spec=data))
 
     return to_add
+
+
+def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config):
+    from jupyter_sphinx.ast import cell_output_to_nodes
```
choldgraf, line 203: short docstring so that we know this is basically just mimicking `jupyter-sphinx` w/ some extra logic

myst_nb/parser.py
```diff
@@ -112,6 +113,33 @@ def parse_block(src, start_line):
                 dup_ref["fixed"] = True
         return tokens
 
+    def parse_code_cell(cell, start_line):
```
chrisjsewell, line 116: move this to a separate function, add documentation, add type annotations

myst_nb/transform.py
chrisjsewell: add documentation, add type annotations

tests/test_glue/test_parser.xml
```diff
@@ -74,6 +78,7 @@
                     glue("key_plt", plt.gcf(), display=False)
             <CellOutputNode classes="cell_output">
                 <image candidates="{'*': '_build/jupyter_execute/with_glue_5_1.png'}" uri="_build/jupyter_execute/with_glue_5_1.png">
+                <image candidates="{'*': '_build/jupyter_execute/with_glue_5_1.png'}" uri="_build/jupyter_execute/with_glue_5_1.png">
```
chrisjsewell, line 81: same as above, why are outputs being duplicated?

tests/test_glue/test_parser.xml
```diff
@@ -16,6 +16,10 @@
                     'text1'
                 <literal_block classes="output text_plain" language="none" xml:space="preserve">
                     3.14159
+                <literal_block classes="output text_plain" language="none" xml:space="preserve">
+                    'text1'
+                <literal_block classes="output text_plain" language="none" xml:space="preserve">
+                    3.14159
```
chrisjsewell, line 22: why are outputs being duplicated?

myst_nb/transform.py
chrisjsewell: also we should consider if/how this can be upstreamed

myst_nb/parser.py
tonyfast: what this is saying is to split code cells on markdown displays. it is possible that there is a text/html or widget following a markdown output and we want to process them.

tonyfast: L136 is the markdown cell you are referring to. the new code cell is the output content potentially following the markdown.

myst_nb/parser.py
tonyfast: do y'all have type checking tests i could run?

i don't think that this function can be moved to a separate function. i am following the pattern used on parse_block and parse-code-cell is invoking parse-block.

myst_nb/parser.py
chrisjsewell: > it is possible that there is a text/html or widget following a markdown output and we want to process them.

this behaviour should be "codified" into a test(s), to demonstrate the behaviour, and check that it is working as expected

tests/test_glue/test_parser.xml
tonyfast: it might have something to do with jupyter cache 🤷🏻

myst_nb/parser.py
tonyfast: where is a good place to put that test?

chris, could you please try to help me along with as i make these first contributions.  these projects have a lot of content and i need help getting familiar. i'd like to help y'all and i could use a little hand holding.

myst_nb/parser.py
chrisjsewell: > do y'all have type checking tests i could run

There is nothing automated setup as yet, but it would be good to add mypy (#204)

> i don't think that this function can be moved to a separate function

The problem is this function is getting quite big and complex. It would be ideal if it could be split in some way

myst_nb/parser.py
chrisjsewell: >  i'd like to help y'all and i could use a little hand holding.

I'll try to help where possible, but unfortunately I don't have a large amount of time to devote to this at present. I appreciate that this is your first contribution, and that's great and I encourage it. 
But at the same time this is not a trivial piece of code being added, and I want to make sure that it is not going to introduce new bugs for all the other users, which it appears to do at the moment.
This first round of review is about identifying potential issues, then we can work on fixing them.

akhmerov:
I can certainly anticipate uses for markdown generated by code, (e.g. some library might have a nice markdown representation of its objects). At the same time, things like e.g. automatically listing the team members from @choldgraf example actually seem to be better done externally.

This is because all the code outputs must be contained within the cell output directive, which in turn makes them render differently, and makes it hard to embed those into the AST on the same level as the rest of the document.

myst_nb/parser.py
tonyfast: could you suggest where you would like this test. is it a standalone test? does it go with an existing test? what bugs am i testing for? 

i'm neurodivergent and i am not good at implicitly understanding what people mean. if you're too busy then maybe i could walk through the review process with another contributor?

chrisjsewell: > it is possible that there is a text/html or widget following a markdown output and we want to process them.

Add a notebook in the tests folder, that has a text/html or widget following a markdown output, then copy one of the other tests to run it and test that the output is as expected 😄 

Then add to the notebook additional "scenarios" of different markdown content, and different cell outputs (markdown then stdout, stdout then markdown, ...)

I'm sure @choldgraf can guide through the rest

myst_nb/parser.py
```diff
@@ -142,15 +143,7 @@ def parse_block(src, start_line):
...
-            block_tokens.append(
-                Token(
-                    "nb_code_cell",
-                    "",
-                    0,
-                    meta={"cell": nb_cell},
-                    map=[start_line, start_line],
-                )
-            )
+            block_tokens.extend(_parse_code_cell(nb_cell, start_line, parse_block))
```
choldgraf, line 146: so here, could we just keep the old code the way it was, and then add a function that checks the outputs for markdown and, if there, appends it to `block_tokens` as a new markdown cell following the logic here: https://github.com/executablebooks/MyST-NB/pull/198/files#diff-a5ed38466bac00edd65797adb0d40849L136?

e.g. something like:

```python
remove_ixs = []
extra_markdown = []
for ii, output in enumerate(cell.outputs):
  if output.get("output_type") == "display_data" & "text/markdown" in output.get("data", []):
    extra_markdown.append(output["data"])
    remove_ixs.append(ii)
for ii in remove_ixs[::-1]:
    cell.outputs.pop(ii)
for markdown_cell in extra_markdown:
    block_tokens.extend(parse_block(markdown_cell , start_line))	
```

tests/test_glue/test_parser.xml
tonyfast: there was duplicate scrapbook metadata causing the output to show twice. this is resolved by popping scrapbook from the new metadata.

tests/test_glue/test_parser.xml
tonyfast: scrapbook metadata

myst_nb/parser.py
tonyfast: I don't know how to get this into a transformer. every which way i tried, i could never use a toctree. having that parse_block function scoped in the parser seems to be collecting headings.

tests/test_glue/test_parser.xml
tonyfast: scrapbook

myst_nb/parser.py
tonyfast: i split the function out of scope as a private function.

myst_nb/parser.py
tonyfast: i may be misunderstanding the comment.

chrisjsewell:
> This is because all the code outputs must be contained within the cell output directive, which in turn makes them render differently, and makes it hard to embed those into the AST on the same level as the rest of the document.

Then question here is then do we always want this to be the case?
It would be perfectly possible to inject the markdown output directly in to the AST since, with myst-nb we execute notebooks before parsing.
Potentially there would be a flag in the sphinx options to turn this on/off.

chrisjsewell:
I'm going to close this as fixed with: https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#markdown 😄

Issue #179, opened 2020-05-19
Make output priority lists configurable


chrisjsewell:
Looking over the code again; it looks like we "hard-code" in the priority lists: https://github.com/executablebooks/MyST-NB/blob/afba1fe9ccae394e9b13621e113ab96a5266c6fe/myst_nb/transform.py#L18

As opposed to in jupyter-sphinx, where it is a configuration variable (albeit only allowing for a single list and not per builder): https://github.com/jupyter/jupyter-sphinx/blob/6a0df8575fd1fc76e1258aab300ec415eee92569/jupyter_sphinx/__init__.py#L155

I think the priority lists should be made configurable, to cope with use cases like this; whereby in general SVG should not be in the priority list for latex, since it will fail, but then you may want to add it in if you are also using an extension that makes it compatible.

_Originally posted by @chrisjsewell in https://github.com/executablebooks/MyST-NB/pull/176#issuecomment-629888036_

chrisjsewell:
closed in v0.9.0

PR #202, opened 2020-06-16
add more readthedocs builders

tonyfast:
This pull request adds a few more sphinx builder names to the render priority.

The new names are derived from https://github.com/readthedocs/readthedocs-sphinx-ext/blob/fb7b5069b422e715346a501b746dc5cb05512977/readthedocs_ext/readthedocs.py#L45 and https://github.com/readthedocs/readthedocs-sphinx-ext/blob/master/readthedocs_ext/readthedocs.py#L83

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,4 @@ repos:
     rev: stable
     hooks:
     - id: black
+      args: [--l=88]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,4 +24,3 @@ repos:
     rev: stable
     hooks:
     - id: black
-      args: [--l=88]
--- a/myst_nb/transform.py
+++ b/myst_nb/transform.py
@@ -51,11 +51,11 @@ class CellOutputsToNodes(SphinxTransform):
             outputs = cell.get("outputs", [])
             if node.get("inline", False):
                 output_nodes = cell_output_to_nodes_inline(
-                    outputs, RENDER_PRIORITY[builder], True, output_dir, None
+                    outputs, RENDER_PRIORITY.get(builder, 'html'), True, output_dir, None
                 )
             else:
                 output_nodes = cell_output_to_nodes(
-                    outputs, RENDER_PRIORITY[builder], True, output_dir, None
+                    outputs, RENDER_PRIORITY.get(builder, 'html'), True, output_dir, None
                 )
             # TODO add warning if output_nodes is empty
             node.replace_self(output_nodes)
--- a/myst_nb/transform.py
+++ b/myst_nb/transform.py
@@ -31,7 +31,11 @@ RENDER_PRIORITY = {
 }
 RENDER_PRIORITY["readthedocs"] = RENDER_PRIORITY["html"]
 RENDER_PRIORITY["singlehtml"] = RENDER_PRIORITY["html"]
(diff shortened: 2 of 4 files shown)


choldgraf:
ah right - this will allow different assets to be built on RTD then?

Also could you double-check the code style? (or just install `pre-commit` in the repo so it checks for code style)

.pre-commit-config.yaml
```diff
@@ -24,3 +24,4 @@ repos:
     rev: stable
     hooks:
     - id: black
+      args: [--l=88]
```
chrisjsewell, line 27: this is already the default for black though?

myst_nb/transform.py
```diff
@@ -47,11 +51,11 @@ def apply(self):
             outputs = cell.get("outputs", [])
             if node.get("inline", False):
                 output_nodes = cell_output_to_nodes_inline(
-                    outputs, RENDER_PRIORITY[builder], True, output_dir, None
+                    outputs, RENDER_PRIORITY.get(builder, 'html'), True, output_dir, None
```
chrisjsewell, line 54: Please also add a sphinx warning towards the start of this transform; to indicate that no render priority is available for the given builder

tonyfast: can you point me to what a sphinx warning looks like?

tonyfast:
i had to add the fallback feature when readthedocs tried to create an epub.

chrisjsewell:
> i had to add the fallback feature when readthedocs tried to create an epub.

Shouldn't there then be a separate build priority for epub, rather than defaulting to html?

myst_nb/transform.py
chrisjsewell: https://www.sphinx-doc.org/en/master/extdev/logging.html#sphinx.util.logging.SphinxLoggerAdapter.warning

tonyfast:
probably, but i just wanted the shit to build and i wasn't sure what other edge cases i was going to run into.

myst_nb/transform.py
```diff
@@ -31,7 +31,12 @@
 }
 RENDER_PRIORITY["readthedocs"] = RENDER_PRIORITY["html"]
 RENDER_PRIORITY["singlehtml"] = RENDER_PRIORITY["html"]
+RENDER_PRIORITY["dirhtml"] = RENDER_PRIORITY["html"]
 RENDER_PRIORITY["linkcheck"] = RENDER_PRIORITY["html"]
+RENDER_PRIORITY["readthedocsdirhtml"] = RENDER_PRIORITY["html"]
+RENDER_PRIORITY["readthedocssinglehtml"] = RENDER_PRIORITY["html"]
+RENDER_PRIORITY["readthedocssinglehtmllocalmedia"] = RENDER_PRIORITY["html"]
+RENDER_PRIORITY["epub"] = RENDER_PRIORITY["html"]
```
chrisjsewell, line 39: @choldgraf @mmcky can you think of any differences in priority between epub and html?

Note this is related to #179
Also I think probably at some point (maybe not in this PR) we just have a list of HTML aliases, rather than this dict copying

choldgraf: None that I can think of, but I must admit I don't understand epub super well. I'd be fine with using the pattern we've already got here, and I agree w/ @chrisjsewell that we should just have a list of aliases rather than copying

choldgraf:
thanks for the patch @tonyfast !! 🎉

PR #203, opened 2020-06-17
fix windows utf8 encoding  part 3/3

phaustin:
see https://github.com/executablebooks/jupyter-book/pull/719

### DIFF
--- a/myst_nb/cache.py
+++ b/myst_nb/cache.py
@@ -221,7 +221,7 @@ def _read_nb_output_cells(source_path, jupyter_execute_notebooks):
         and jupyter_execute_notebooks == "auto"
         and "ipynb" in ext
     ):
-        with open(source_path, "r", enconding="utf8") as f:
+        with open(source_path, "r", encoding="utf8") as f:
             ntbk = nbf.read(f, as_version=4)
             has_outputs = all(
                 len(cell.outputs) != 0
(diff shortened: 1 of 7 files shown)


phaustin:
@choldgraf -- I reverted one change to get the file_regression test to pass -- the open statement
in the notebook artifact should also specify encoding="utf8":
i.e. [this commit](https://raw.githubusercontent.com/phaustin/MyST-NB/0c48c6ecf37f44aa0e44b057044ceb50f3fe34ee/tests/notebooks/basic_relative.ipynb)

choldgraf:
@phaustin re: the file regression, I'm not quite sure what you mean by reverted. do you mean there's one part in this codebase that doesn't use utf-8 now?

phaustin:
I mean that if I add encoding="utf8" to the ipynb version of [this file](https://github.com/phaustin/MyST-NB/blob/cd3687b9321ad1385040f1fc2a2a5ec0351917e1/tests/notebooks/basic_relative.md#a-title)
pytest fails with a "file regression" error.

choldgraf:
could you post the error?

phaustin:
reran with the same command line arguments as the CI, and all tests pass locally on my mac, I think it was a timeout issue, so I think this is ready to merge. On my local windows there are still about half a dozen pytest failures, but several of those are related to the windows path problem.  I've isolated that issue to  app.env.doc2path in cache.py (and also in jupyter-book/toc.py) and we can retest once I've pushed those PRs.

choldgraf:
ok cool! let's merge this in and see how things go :-) thanks for the improvement!

PR #209, opened 2020-07-04
no longer activating jupyter sphinx

choldgraf:
let's see what tests this breaks :-)


choldgraf:
I think we're using the jupyter-sphinx extension for more than I expected under the hood! so for now seems reasonable to keep it

PR #213, opened 2020-07-15
adding more explanation of myst notebooks

choldgraf:
This is an attempt at clarifying a few things that were getting intermixed in our documentation, to try and resolve some conceptual issues that arose in #210 . It does the following:

* Creates a top-level page about MyST Markdown Notebooks, with links to other relevant documentation
* Adds more information about MyST Notebooks like the "spec" for a minimal MyST Notebook, with lots of cross-refs to other documentation for how to use MyST Notebooks in other tools
* Re-focuses the `Use and Configure` section to explicitly be about the Sphinx extension.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,12 @@ master_doc = "index"
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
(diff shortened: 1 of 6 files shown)


choldgraf:
gonna self-merge this since it's just a docs improvement - happy to iterate on things if people think of subsequent changes to make!

chrisjsewell:
Yep looks good to me 👍

PR #218, opened 2020-08-04
✨ Add configuration for traceback in stderr

chrisjsewell:


chrisjsewell:
I f we are using RTD for CLI, we can probably remove circleci (now that https://github.com/readthedocs/readthedocs.org/issues/7240 is closed)
weird that circleci passes and RTD does not, clearly something in the environment

chrisjsewell:
ok weird, now everything passes, but the outputs are still not there.
If you download the notebook from the footnote in: https://myst-nb--218.org.readthedocs.build/en/218/use/hiding.html, the matplotlib images have actually been created (as `image/png` MIMEs) but these are not making their way into the HTML output

chrisjsewell:
Also @choldgraf can you add https://readthedocs.org/profiles/ebp-bot as a maintainer of the RTD

choldgraf:
@chrisjsewell I added the bot!

chrisjsewell:
I don't think this fully fixes the original issue, but I am going to merge and open a new PR

Issue #99, opened 2020-03-22
Ensure Math Is Parsed/Rendered Consistently


phaustin:
The top two equations [here](https://github.com/phaustin/MyST-NB/blob/tex_format/docs/examples/austin_issue/lab8/lab8_snip.ipynb) succeed with nbsphinx but not myst_nb.    Reference PR #98

chrisjsewell:
FYI @phaustin this shoud hopefully be fixed in ExecutableBookProject/MyST-Parser#123

phaustin:
Hi @chrisjsewell -- trying this, but I'm getting the following error when I try to do `make html` in docs:

first I do 
```
pip install myst-nb==0.6.0a1
pip install myst-parser[sphinx]==0.8.0a3
```
then checkout current tip

```
% git log -4
commit f07f2253297df60a33fb23b32c71af0b2c819016 (HEAD -> master, origin/master, origin/HEAD)
Author: AakashGC <aakashguptachoudhury@gmail.com>
Date:   Wed Apr 1 06:35:29 2020 +1100

    Documentation for unit testing (#112)
```

```
% make html
Running Sphinx v2.4.4
WARNING: while setting up extension jupyter_sphinx: extension 'jupyter_sphinx' has no setup() function; is it really a Sphinx extension module?
Adding copy buttons to code blocks...
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 13 source files that are out of date
updating environment: Executing: /Users/phil/repos/MyST-NB/docs/use/interactive.ipynb
Execution Succeeded: /Users/phil/repos/MyST-NB/docs/use/interactive.ipynb
Executing: /Users/phil/repos/MyST-NB/docs/use/glue.ipynb
Execution Succeeded: /Users/phil/repos/MyST-NB/docs/use/glue.ipynb
Executing: /Users/phil/repos/MyST-NB/docs/use/orphaned_nb.ipynb
Execution Succeeded: /Users/phil/repos/MyST-NB/docs/use/orphaned_nb.ipynb
Executing: /Users/phil/repos/MyST-NB/docs/use/hiding.ipynb
Execution Succeeded: /Users/phil/repos/MyST-NB/docs/use/hiding.ipynb
Executing: /Users/phil/repos/MyST-NB/docs/use/basic.ipynb
Execution Succeeded: /Users/phil/repos/MyST-NB/docs/use/basic.ipynb
[...]

PR #226, opened 2020-08-12
⬆️ UPGRADE: myst-parser v0.12 & sphinx v3

chrisjsewell:


chrisjsewell:
This PR now removes the MathJax overrides introduced in #121 (as quick fix for #99)

As discussed previously (e.g. https://github.com/executablebooks/MyST-NB/issues/126#issuecomment-622935821) these overrides are bad, because (a) they only work for mathjax (e.g. not for katex or imgmath) and (b) they override any configuration set *via* sphinx (see [confval-mathjax_config](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_config)).

@phaustin can you try this out and/or provide some examples of math where there was previously an issue (most of the links you provided in #99 now appear to be dead)

chrisjsewell:
See this page where I have updated the examples: https://myst-nb--226.org.readthedocs.build/en/226/examples/basic.html

chrisjsewell:
You may also *note* the new admonition syntax:

:::{note}
This is a note admonition that **will** render content correctly here on GitHub and in notebooks etc

:::

chrisjsewell:
Currently for these admonitions and the latex equations, users have to specifically set in conf.py

```python
myst_admonition_enable = True
myst_amsmath_enable = True
```

I'm not sure if they should be enabled by default?

choldgraf:
This is really nice - two quick thoughts:

**re: making this default**

I can definitely see how "things that behave like directives but that have markdown inside" could be bracketed with `:::` instead of ` ``` `. At the least, it makes things look nicer inside of Jupyter / non-MyST interfaces.

What's the process by which we should decide if this is supported *out of the box* in MyST? I'd think something like
- Make it possible, but not default for a while, capture usecases and experiences
  - This could be possible in Jupyter Book as well as an "optional" flag
- Make an official proposal (perhaps similar to [a JEP](https://jupyter.org/enhancement-proposals/jupyter-enhancement-proposal-guidelines/jupyter-enhancement-proposal-guidelines.html) but perhaps more lightweight) for why it's a good idea to extend the MyST syntax in this new way
- Solicit feedback from the community etc on that proposal
- EBP core team members vote on whether to accept the proposal

I think that full process is less important in the early days, but will become more important over time if we want MyST to be a proper "standard" that others use. 

**re: your admonition syntax**

Where does the `:::{admonition,note}` syntax come from? Is that a MyST-specific syntax that maps to

```
:::{admonition}
:class: note

:::
```

?

myst_nb/_static/mystnb.js
```diff
@@ -1,44 +0,0 @@
-// Initialize MathJax with the notebook config
```
chrisjsewell, line 1: @choldgraf is there anything here you could see as being an issue because we've removed it

choldgraf: Hmmm I don't _think_ so as long as we have the same behavior for catching math syntax delimeters. The notebook mathjax config is actually pretty non-specific and isn't strongly documented anywhere, so we have a little bit of wiggle room in general, we just don't want to surprise people with MyST's behavior I think. Let's try it out and see if people start complaining :-)

chrisjsewell:
> Where does the :::{admonition,note} syntax come from? Is that a MyST-specific syntax that maps to

As explained [here](https://markdown-it-py.readthedocs.io/en/latest/plugins.html) the underlining implementation is the markdown-it-container plugin, which are also equivalent to [pandoc divs](https://pandoc.org/MANUAL.html#divs-and-spans)
 
The `{}` were mainly chosen to be equivalent to the normal myst directives ```` ```{admonition} ````

As mentioned in [the docs](https://myst-parser.readthedocs.io/en/latest/using/syntax.html#admonition-directives-special-syntax-optional), because the content is parsed as Markdown, it is not trivial to retrieve actual options (like `:class: note`) from inside the container, so having additional classes (which are really the only option necessary for admontions) inside the `{}` was the simplest solution I could think of

chrisjsewell:
> At the least, it makes things look nicer inside of Jupyter / non-MyST interfaces.

Well thats the only thing it does really lol. But I think that's really important for adoption by joe public, and would certainly like to transition this to be the *de facto* standard in our documentation

> What's the process by which we should decide...

yep something like that sounds good.

choldgraf:
Thanks for those links to the docs. Do we imagine keeping a strict scoping to "container + classes" for the `:::` syntax? If so, I feel like we could also relax some of the restrictions around `{}`. That way we could suggest:

- ` ```language ` implies raw text that should be marked-up by a `<language>` highlighter
- ` ```{thing} ` implies interpreted text that will be interpreted by some kind of program and has variable outputs and behavior
- `:::class1,class2` is a way to tag blocks of markdown with arbitrary labels, and some of those labels may have special meaning in different build systems
  - On this one, there will be the follow up question of "OK so which labels are 'special' and who decides this?"
  - There is also the special-case of `::class,class <title>` for admonitions. I almost wonder if we should make `<title>` true for *all* `:::` blocks, and then find ways to "make it work" under the hood

(apologies if some of this is obvious to you, I'm just thinking out loud to make sure I'm grokking things properly)

chrisjsewell:
> `:::label1,label2` is a way to tag blocks of markdown with arbitrary labels

well it would be more `:::thing,class1,class2`, i.e. it is the first label that defines how the block is interpreted.

> OK so which labels are 'special' and who decides this?

The ones we dictate lol!
Because they have to be specifically programmed into myst-parser (at least currently), as opposed to directives where we "delegate" responsibility for converting the text to AST to the directive.

chrisjsewell:
> I almost wonder if we should make <title> true for all ::: blocks

Well at present you can write:

```
:::{note} title
body
:::
```

it will just ignore the title

chrisjsewell:
it won't alway be a <title> though, as with directive, <argument> is more apt.

The other one I'm probably going to implement soon is https://github.com/executablebooks/MyST-Parser/issues/208#issuecomment-672806605:

```
:::{figure}
<img src="pic_trulli.jpg" alt="Trulli" width="200px>
This **is** the caption
:::
```

because I think thats another one that would be quite generally useful.


> are we slowly just converging on "MyST markdown is just Pandoc markdown"

Well in a way yes, because we are trying to use "standard" markdown extension syntaxes, which often correlate with pandoc. Although moreso its utiling existing markdown-it plugins: https://markdown-it-py.readthedocs.io/en/latest/plugins.html
and obviously pandoc itself can't create docutils/sphinx AST 😄

chrisjsewell:
New in 0.11.2; see https://myst-parser.readthedocs.io/en/latest/using/syntax.html#images for using HTML image tags in Markdown (e.g. ``) that get handled correctly by sphinx 😄

chrisjsewell:
So I'm good to go with this PR; just waiting on some possible testing of the LaTeX/dollar equations by whomever (also cc @jstac and @mmcky).

mmcky:
thanks for the ping @chrisjsewell -- I could add some `math` test cases to the `test` suite tomorrow.

chrisjsewell:
> I could add some math test cases to the test suite tomorrow.

cheers @mmcky but you probably don;t need to add any more test cases here, since they are already done in markdown-it-py and myst-parser, and tests would capture the mathjax side of things. Just have a play around with building the docs here I guess and see if any math you add (latex or dollar) works as expected

mmcky:
hey @chrisjsewell just did some quick testing -- the math is looking pretty great. I added the following
[math.md.txt](https://github.com/executablebooks/MyST-NB/files/5067199/math.md.txt) (remove `txt`) and the output is as I expected. I can do some more extended testing tomorrow. 

Screen Shot 2020-08-13 at 2 54 19 pm

chrisjsewell:
Another thing to check would be what now happens with this issue: https://github.com/executablebooks/jupyter-book/issues/750

mmcky:
thanks for those references @chrisjsewell -- I can do some more extended testing today. 

re: issue -- inline escapes aren't resolved.

chrisjsewell:
Cheers 😀

chrisjsewell:
> re: issue -- inline escapes aren't resolved.

Ok, I will look to fix this in https://github.com/executablebooks/markdown-it-py/issues/25, but obviously that will be after this PR

PR #241, opened 2020-08-21
🐛 FIX: Update of `nb-exec-table`s

chrisjsewell:
Also

- Use custom `SphinxError` for invalid configuration values
- fixed bug when code cell language of failed notebooks was being added as None
- Sort table rows by docname

### DIFF
--- a/myst_nb/__init__.py
+++ b/myst_nb/__init__.py
@@ -5,6 +5,7 @@ from pathlib import Path
 
 from docutils import nodes
 from sphinx.application import Sphinx
+from sphinx.errors import SphinxError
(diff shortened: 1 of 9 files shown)


chrisjsewell:
FYI @mmcky @AakashGfude I had to make a few changes to the execution data capture, to ensure the data/tables are updated correctly when re-building docs.

I didn't actually realise that only certain documents are re-run through post-transforms (see https://github.com/sphinx-doc/sphinx/pull/8140)

mmcky:
thanks for the ping @chrisjsewell -- good to know.

PR #243, opened 2020-08-23
Improve Notebook Output Rendering

chrisjsewell:


mgeier:
If you need a few test cases for ANSI coloring, feel free to have a look at https://nbsphinx.readthedocs.io/code-cells.html#ANSI-Colors

I've implemented the ANSI support in `nbconvert`, `notebook` and `jupyterlab`, so I know a few of the pitfalls.
If you need pointers, please let me know.

chrisjsewell:
Thanks @mgeier!
I added your 8-bit ANSI into the docs, and it looks to be working well 😄 https://myst-nb--243.org.readthedocs.build/en/243/use/formatting_outputs.html#ansi-outputs
I'll leave 256-but ANSI for another time/PR though (or feel free to give it a go 😬)

chrisjsewell:
Hey @mmcky and @AakashGfude (and I see you lurking @choldgraf 👀 😆) I have a few more bits to add on this, but the main restructuring and features are there if you want to have a look and give any thoughts.
The commit messages and this page should explain the main changes/additions: https://myst-nb--243.org.readthedocs.build/en/243/use/formatting_outputs.html

chrisjsewell:
Oh and also @akhmerov; a) because I'm sure it'll be of interest, but also b) because you can now close https://github.com/jupyter/jupyter-sphinx/pull/130 however you see fit, because I don't even use `cell_output_to_nodes` any more lol

akhmerov:
@chrisjsewell thanks for the heads up, this is indeed very informative. At a glance `render_outputs.py` seems like it could as well be in jupyter_sphinx. Did you consider that option?

chrisjsewell:
Well we can have a look and think about that at a later date, but for now (a) its too much complication, in terms of cross-repo testing, documentation, reviewing, etc, but also (b) it has dependencies on myst-parser, myst-nb configuration values, and relies on the use of cell metadata, which jupyter-sphinx's directives might have difficulty handling.

So I'm open to the possibility, but I'm afraid I haven't got time to figure out how 😬

akhmerov:
Sounds reasonable. As a quick note: factoring out myst-specifc parts would amount to making an entry point for the myst-markdown renderer.

chrisjsewell:
> Passing render state via cell tags seems alright.

Well, its going to be more than just tags...

akhmerov:
Really? That seemed to be the only context where `metadata` occurs in `render_outputs.py`.

chrisjsewell:
> Really? That seemed to be the only context where `metadata` occurs in `render_outputs.py`.

Thats now, in a few hours, maybe not lol

chrisjsewell:
@akhmerov now this is what I'm talking about 😄  https://myst-nb--243.org.readthedocs.build/en/243/use/formatting_outputs.html#images

docs/use/formatting_outputs.md
```diff
@@ -0,0 +1,218 @@
...
+- **align**: "top", "middle", "bottom", "left", "center", or "right"
+- **classes**: space separated strings
+- **alt**: string
+
+Units of length are: 'em', 'ex', 'px', 'in', 'cm', 'mm', 'pt', 'pc'
+
+We can also set a caption (which is rendered as [CommonMark](https://commonmark.org/)) and name, by which to reference the figure:
+
+````md
+```{code-cell} ipython3
```
mmcky, line 91: @chrisjsewell what happens if the code in this `code-block` doesn't produce an `image`? Do we throw an error if the returned `mime` type of the `output` block isn't compatible with the `image` directive?

chrisjsewell: Then the `render_image` method is never called, and the metadata just isn't used,
i.e. the metadata is accessed (and validated) "lazily" when it is required

chrisjsewell: If it is invalid, for example I give a bad width here, then it produces a warning, and since this a text-based notebook, that warning will point to the correct line of the document (the first one of the code cell)

````md
```{code-cell} ipython3
---
myst:
  image:
    width: abc
---
from IPython.display import Image
Image("images/fun-fish.png")
```
````

```
/Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/use/formatting_outputs.md:110: WARNING: output render: Invalid image attribute: (key: 'width'; value: abc)
not a positive measure of one of the following units:
"em" "ex" "px" "in" "cm" "mm" "pt" "pc" "%"
```

and also, because the image is no longer created:

```
/Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/use/formatting_outputs.md:124: WARNING: 'myst' reference target not found: fun-fish
```

**Review by mmcky — COMMENTED** [review]
thanks @chrisjsewell. It is neat to be able to apply styles and formatting on `images` that are generated by code and the `ansi` support is timely on the `LaTeX` side. 

For `ansi` support -- I came across the [ansifilter](https://gitlab.com/saalen/ansifilter) project this morning that may be useful in converting to various formats. However I don't think it is directly importable to python that I can tell. 

For `LaTeX` output -- from what I can tell reading the code you parse `ansi` codes and return a new `token` object for inclusion into the `AST` so we just need to support those new node types when writing the `tex` (or skip those elements for stripping output of `ansi` codes).

File: docs/use/formatting_outputs.md (hunk 1 above)
Comment (mmcky) [reply docs/use/formatting_outputs.md:91]: I see -- thanks @chrisjsewell

File: docs/use/formatting_outputs.md [hunk 2]
```diff
@@ -0,0 +1,218 @@
...
+By default, the standard output/error streams and text/plain MIME outputs may contain ANSI escape sequences to change the text and background colors.
+
+```{code-cell} ipython3
+import sys
+print("BEWARE: \x1b[1;33;41mugly colors\x1b[m!", file=sys.stderr)
+print("AB\x1b[43mCD\x1b[35mEF\x1b[1mGH\x1b[4mIJ\x1b[7m"
+      "KL\x1b[49mMN\x1b[39mOP\x1b[22mQR\x1b[24mST\x1b[27mUV")
+```
+
+This uses the built-in {py:class}`~myst_nb.ansi_lexer.AnsiColorLexer` [pygments lexer](https://pygments.org/).
```
mmcky, line 156: @chrisjsewell is there a reason we don't just import this from `pygments`. Avoiding a hard requirement?

chrisjsewell:
> that may be useful in converting to various formats

Well, fingers crossed, that should already happen, because we are using [pygments](https://pygments.org/) to lex it to tokens, then sphinx chooses the appropriate formatter to convert it: https://pygments.org/docs/formatters/

docs/use/formatting_outputs.md
chrisjsewell: Because its not part of pygments, there is no built-in ANSI lexer, this is a bespoke one

chrisjsewell: It appears to work well for the 8-bit ansi though, perhaps some finessing of the CSS.
Then, as it says in the docs, we can get round to the adding 256-bit support as and when necessary

mmcky: Oh OK -- sorry had assumed it was adapted from `pygments` re: header comment. Makes sense.

mmcky: Is this something we should contribute upstream to `pygments` at some point?

chrisjsewell: Yeh possibly; just wanted to get everything working first, and then we can look to tidy up

chrisjsewell:
@mmcky has given i the thumbs up for "working" (or at least not breaking) LaTeX, so I'm going to merge this monstrosity!

docs/use/formatting_outputs.md
```diff
@@ -0,0 +1,218 @@
...
+You can change the lexer used in the `conf.py`, for example to turn off lexing:
+
+```python
+nb_render_text_lexer = "none"
```
mmcky, line 160: @chrisjsewell one last question. Does this option in `conf.py` just control `ansi`?

chrisjsewell: yep, it controls what pygments lexer is applied to plain text outputs (including stdout/stderr)

mmcky: ok - so it does more than just `ansi`. Was just thinking it might be better named `nb_render_text_ansi` but it is more generally a control over the lexer used which contains more than just `ansi` parsing.

mgeier:
> I added your 8-bit ANSI into the docs, and it looks to be working well

image

It doesn't seem to support "underline" and "invert", and there don't seem to be any "intense" colors when switching to "bold".

image


> I'll leave 256-but ANSI for another time/PR though (or feel free to give it a go)

I've already implemented this in three different places (in two languages), and I don't currently have the energy to do it yet again.

Do you even have to use Pygments? Why not just use `nbconvert`'s implementation?

For reference, the `nbconvert` implementation is there: https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/ansi.py

PR #261, opened 2020-08-31
🐛 FIX: remove cell background as none

melaniewalsh:
This suggestion fixes the issue with Jupyter Book that I raised last week, which was that the toolbars for Bokeh data visualizations are not showing up. Removing cell backgrounds as none allows them to show up correctly.
https://github.com/executablebooks/jupyter-book/issues/905

### DIFF
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -7,7 +7,6 @@ div.container.cell {
 /* Removing all background formatting so we can control at the div level */
 .cell_input div.highlight, .cell_input pre, .cell_output .output * {
   border: none;
-  background: none;
   background-color: transparent;
   box-shadow: none;
 }


choldgraf:
nice! thanks so much for the patch 👍 ✨ 

I suspect this might result in unexpected behavior for some other cell outputs, though the ones we demo all look fine to me, so I think this is a good improvement (and forcing all cell outputs to have background of `none` was maybe heavy-handed)

chrisjsewell:
Cheers. 
So no other side-effects we can think of?
It's always a bit tricky to review css changes, but I guess if the docs look ok.

choldgraf:
yeah they look OK to me and as you say, hard to know otherwise...I have no idea how to test for regressions in a principled way w/ CSS.

Thanks so much @melaniewalsh for the improvement!!!

PR #271, opened 2020-10-12
🐛 FIX: empty myst file read

chrisjsewell:


filippo82 (approved):
As simple as it gets 👍

PR #279, opened 2020-11-21
MAINTAIN: Fix homepage link

rkdarst:
- I just noticed that homepage link did not work.  I think the change
  is self-explinatory and navigating to the web to fill a possible PR
  template isn't worth the time.
- Review: none probably necessary


choldgraf:
ah thanks very much for noticing this!

PR #280, opened 2020-11-21
🔧 MAINTAIN: Update Python classifiers in setup

rkdarst:
- Classifiers reported working in Python 3.3-3.6, but we currently
  only support >=3.6.  Update classifiers to 3.6-3.9 to match this.
- Review: general check, but also think if there is a better practice
  than keeping the same info twice.


choldgraf:
looks great - thanks for this improvement ✨

chrisjsewell:
Yep cheers 👍

PR #281, opened 2020-11-21
MAINTAIN: fixing docs build

choldgraf:
This fixes some configuration that was throwing an error in the docs


chrisjsewell:
Haha yeh I was going poke you about nerving the last PR without first fixing this 😝

PR #286, opened 2021-01-05
✨ NEW: Add file load option for code-cells

mmcky:
This PR enables the use of a file option for `code-cell` blocks such as:

````
```{code-cell} ipython3
:load: _static/somefile.py
```
````

**Discussion:**

This is a pretty simple way forward to solve #274 at the `myst_nb` level during the conversion from `md` to `ipynb` file. It essentially looks for a `file` option attribute and if one is found it brings the contents of the file in and replaces the body_lines from the parsed `code-block` token. This produces an `ipynb` that has the contents of the file as a code-block. 

The downsides of implementing this feature here are:
1. `myst_nb` provides this feature (rather than core `myst-parser`) so wouldn't be considered part of `myst`. 
2. It is also only implemented for `md` file sources. 

It is my sense that this feature would rarely be used at the `ipynb` layer so this could be considered a nice convenience feature of `myst_nb` but hesitate to cause confusion with other parser features. This small addition would solve our issue of bringing `executable code in from a file` -- but understand if others think this might be better implemented elsewhere. I am open to thoughts and discussion.

~~**Note:** this is not complete as it would need `tests`, `docs`, and some `robustness` additions~~

### CODE REVIEW

**Comment by jstac** [comment]
Very useful and nice simple syntax.

**Comment by choldgraf** [comment]
I think this is a pretty straightforward solution on the myst nb side. I suspect that the main downside is that we are breaking from the “two way lossless conversion to ipynb via jupytext only” constraint that we have been following thus far. This is similar to the “code cells can’t be nested in other things” question as well, I think.

I guess to me the main question is how to gracefully support this “extra” functionality in a way that doesn’t confuse people, or create strong maintenance burden, etc.

**Comment by mmcky** [comment]
> I guess to me the main question is how to gracefully support this “extra” functionality in a way that doesn’t confuse people

I agree, what I propose is the following:

1. Use `myst-nb` to support this as a `jupyter-book` feature for `myst:markdown`
2. (possibly) document this feature in the `myst-nb` documentation only given it isn't a formal `myst` syntax feature
3. Open an enhancement ticket for the `myst` specification for future support
4. Deprecate once `MyST` supports the markup natively

**Comment by mmcky** [comment]
> two way lossless conversion to ipynb via jupytext only

this may not be overly broken but will need to dig a bit further. They way I have chosen to update `block_lines` only doesn't actually impact on the `token` object which means `cell metadata` is transferred to the notebook. So in this case the `file` and `path` are added to the cell metadata which `jupytext` should understand (I think).

**Comment by choldgraf** [comment]
@mmcky that's a good point - the resulting ipynb will be two-way lossless, it's just that the ipynb created by jupytext alone (w/o myst nb) will have an empty code cell with the "file" metadata in it. Is that right?

**Comment by mmcky** [comment]
> @mmcky that's a good point - the resulting ipynb will be two-way lossless, it's just that the ipynb created by jupytext alone (w/o myst nb) will have an empty code cell with the "file" metadata in it. Is that right?

@choldgraf I think that is what would happen -- but I will confirm this in the morning :-).

**Comment by mmcky** [comment]
So I had a quick look -- the answer is almost but not quite. I used the following source `md` file:

````md
---
jupytext:
  text_representation:
    extension: .md
    format_name: myst
kernelspec:
  display_name: Python 3
  language: python
  name: python3
---

# Simple Example

```{code-cell} python3
:file: _static/test.py
```
````

which builds the `ipynb` as expected

````json
{
 "cells": [
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "# Simple Example"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {
    "file": "_static/test.py"
   },
   "outputs": [],
   "source": []
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 4
}
````

listing the `file` option as `metadata`. 

However the return to `md` isn't formatted quite right using `jupytext==1.8.2` where the `code-cell` is returned as

````md
```python file="_static/test.py"

```
````

**Comment by mmcky** [comment]
@chrisjsewell within the test framework do you know how I can copy a support file to the `var` location that the test are executed in?

**Comment by AakashGfude** [comment]
@mmcky I reckon we can copy the support file explicitly from the source to the "temp srcdir" . Implicit copying does not sound feasible/practical to me.
Code can be something like:
```
from sphinx.util.fileutil import copy_asset_file

def test_codecell_file(sphinx_run, file_regression, check_nbs, get_test_path):
    asset_path = get_test_path('mystnb_codecell_file.py')
    copy_asset_file(str(asset_path), str(sphinx_run.app.srcdir))
```

**Comment by mmcky** [comment]
ok thanks @AakashGfude I will try and add that in today and see if that works.

**Comment by mmcky** [comment]
@AakashGfude is 48129ad what you had in mind?

**Comment by AakashGfude** [comment]
> @AakashGfude is [48129ad](https://github.com/executablebooks/MyST-NB/commit/48129adc49b98a0085ca45f8373e4b9d347b19d5) what you had in mind?

that's right @mmcky. Now, you just need to push `test_codecell_file.ipynb` and `test_codecell_file.xml`, created during the local test run, in this PR. So, that these file_regression check lines https://github.com/executablebooks/MyST-NB/pull/286/files#diff-2f71d7670a4c11fabd3fd5891cdcd501f695c2ee50b53bdf7fbcaf6bf669a8b0R27 do not try to create them during CI run.

**Comment by mmcky** [comment]
thanks @AakashGfude with your help in getting this test infrastructure updated.

**Comment by mmcky** [comment]
@AakashGfude would you mind to review this PR when you have some time. It would be great to get this into the next release of `myst_nb`

**Comment by AakashGfude** [comment]
One case in which we need to issue a warning is when we have both inline code and `:file:` key in a code-cell. Example:
```
```{code-cell} ipython
:file: ../TOCTESTBOOK/mystnb_codecell_file.py
i = 10
print(i)
```


The `file` imported by the :file: key overwrites the inline code at present. We should notify users when this happens.

mmcky:
thanks for this addition @AakashGfude 👍

AakashGfude (dismissed):

docs/use/markdown.md
```diff
@@ -103,6 +103,22 @@ b = "Python code!"
...
+you specify a `file` metadata attribute such as:
+
+````md
+```{code-cell} ipython3
+:file: <path>
+```
+````
+
+```{warning}
+This is a feature provided by `myst_nb` and is **not** part of the core `MyST` markup specification.
```
chrisjsewell, line 119:
```suggestion
This is an experimental feature that is **not** part of the core `MyST` markup specification, and may be removed in the future.
```

docs/use/markdown.md
```diff
@@ -103,6 +103,22 @@ b = "Python code!"
...
 
+**Code from Files:**
+
+`myst_nb` provides a convenience feature for importing executable code into a `{code-cell}`
+from a file. This can be useful when you want to share code between documents. To do this
+you specify a `file` metadata attribute such as:
+
+````md
+```{code-cell} ipython3
+:file: <path>
```
chrisjsewell, line 114:
```suggestion
:myst_file: <path>
```

docs/use/markdown.md
```diff
@@ -103,6 +103,22 @@ b = "Python code!"
 print(f"{a} {b}")
 ```
 
+**Code from Files:**
+
+`myst_nb` provides a convenience feature for importing executable code into a `{code-cell}`
+from a file. This can be useful when you want to share code between documents. To do this
+you specify a `file` metadata attribute such as:
```
chrisjsewell, line 110:
```suggestion
you specify a `myst_file` metadata attribute such as:
```

myst_nb/converter.py
```diff
@@ -188,6 +197,7 @@ def myst_to_notebook(
     :param raw_directive: the name of the directive to search for containing raw cells
     :param add_source_map: add a `source_map` key to the notebook metadata,
         which is a list of the starting source line number for each cell.
+    :param: context: passthrough Sphinx Builder Environment
```
chrisjsewell, line 200: re-write this so you don't pass the sphinx builder to this function, which should be independent of sphinx

myst_nb/converter.py
```diff
@@ -248,6 +258,19 @@ def _flush_markdown(start_line, token, md_metadata):
         if token.type == "fence" and token.info.startswith(code_directive):
             _flush_markdown(md_start_line, token, md_metadata)
             options, body_lines = read_fenced_cell(token, len(notebook.cells), "Code")
+            # Parse :file or file: tags and populate body with contents of file
+            if "file" in options:
```
chrisjsewell, line 262:
```suggestion
            if "myst_file" in options:
```

myst_nb/converter.py
```diff
@@ -248,6 +258,19 @@ def _flush_markdown(start_line, token, md_metadata):
...
+                fl = re.search(r"( .*?\.[\w:]+)", token.content).group(0).lstrip()
+                flpath = Path(context.srcdir + "/" + fl).resolve()
```
chrisjsewell, line 264: this would fail if context was None, and anyway as above this function should not be dependent on sphinx.

myst_nb/converter.py
```diff
@@ -248,6 +258,19 @@ def _flush_markdown(start_line, token, md_metadata):
...
+                if len(body_lines):
+                    LOGGER.warning(
+                        ("content of code-cell is being overwritten by `file`"),
+                        location=(context.srcdir + "/" + context.docname, token.map),
```
chrisjsewell, line 268: passing `token.map` to location is wrong

chrisjsewell (changes requested):

chrisjsewell:
>  I suspect that the main downside is that we are breaking from the “two way lossless conversion to ipynb via jupytext only” constraint that we have been following thus far.

yeh again, for the record, I think this is a bad idea that breaks our pre-agreed design decision. But I don't have time to think of an alternative right now, so I'll approve with the changes I have requested.

mmcky:
> > I suspect that the main downside is that we are breaking from the “two way lossless conversion to ipynb via jupytext only” constraint that we have been following thus far.
> 
> yeh again, for the record, I think this is a bad idea that breaks our pre-agreed design decision. But I don't have time to think of an alternative right now, so I'll approve with the changes I have requested.

thanks @chrisjsewell for your comments and review. I will go through them today. 

As per discussion above I tested this with `jupytext` and the conversion is effectively lossless - but doesn't action the `:file:` import through `jupytext` -- instead it adds it as cell metadata by default -- so we could update `jupytext` to action `:file:` metadata tag.

myst_nb/converter.py
mmcky: @chrisjsewell we need to use the `srcdir` from the `builder environment` to adjust the `path`. Would you prefer I extract `srcdir` and pass that instead of the passing the builder `context`?

docs/use/markdown.md
mmcky: @chrisjsewell not sure why you want this to be named `myst_file` as they are files that contain code such as `py` and `jl` files.

myst_nb/converter.py
mmcky: @chrisjsewell can you please explain why you'd like `myst_file` (as the files usually have no relationship with `myst` files)?

myst_nb/converter.py
mmcky: @AakashGfude can you review this comment and update. Thanks.

myst_nb/converter.py
mmcky: @chrisjsewell I have removed `context` passing and added `file_srcdir` to support the `file` import option.

myst_nb/converter.py
mmcky: @chrisjsewell I have removed context passing and added file_srcdir to support the file import option in 12a11a3

docs/use/markdown.md
mmcky: I updated the error name from `MystFileParsingErorr` to `CodeFileParsingError` to avoid confusion.

myst_nb/converter.py
AakashGfude: hmm. I was trying to get the line numbers for this logger. Could not find another variable that can pinpoint that. @chrisjsewell why is it wrong though? should I just get one entry(the start line) from `token.map` and pass it?

mmcky: thanks @AakashGfude I think adding line number information is a big help to the `warning`.

mmcky:
@chrisjsewell it looks like `tox` doesn't specifically control the version used of `myst_parser`. We had to update this fixture when I committed the change to `docs` page in 32523ca. Just checking that update to the fixture makes sense with you.

myst_nb/converter.py
mmcky: I need to use more than one context based variable so I have renamed the context to `file_context` to indicate it is used by the `:file:` option to `code-cell`. That will make it easier to remove if this get's implemented elsewhere.

myst_nb/converter.py
mmcky: I need to use more than one context based variable so I have renamed the context to file_context to indicate it is used by the :file: option to code-cell in 3207b34. That will make it easier to remove if this get's implemented elsewhere.

mmcky:
@chrisjsewell I have now responded to all your review comments above and put reasons for not accepting some. Let me know if these updates make sense. It would be great to get this feature incorporated so we can finalise our migration projects.

choldgraf:
I think that since we're flagging this feature as strictly experimental, and not part of the core myst-notebook spec, we have a little bit of leeway in changing the implementation as we learn more and see how people use it (even it this means removing the implementation entirely!). As long as we make it clear to others what to expect this is IMO ok. So I'd be +1 on merging in and seeing what kinds of feedback we get.

AakashGfude (dismissed):

chrisjsewell:
> it looks like tox doesn't specifically control the version used of myst_parser. We had to update this fixture when I committed the change to docs page in 32523ca. Just checking that update to the fixture makes sense with you.

yep thats fine; tox uses the version pinning in setup.py, this change comes from moving from myst-parser 0.13.3 to 0.13.5

chrisjsewell:
> I think that since we're flagging this feature as strictly experimental, and not part of the core myst-notebook spec, we have a little bit of leeway in changing the implementation as we learn more and see how people use it (even it this means removing the implementation entirely!). As long as we make it clear to others what to expect this is IMO ok.

yes exactly, once it is in and people starting using it, it will be hard to deprecate

chrisjsewell:
> so we could update jupytext to action :file: metadata tags.

yes I would certainly ask if you can pursue this (or another solution) in jupytext

docs/use/markdown.md
chrisjsewell: my tentativeness here is that `file` is a pretty generic key and on the off-chance anyone already has this key in their metadata it will now break their builds. Is there perhaps something a little less generic; `import_file` or whatever?

myst_nb/converter.py
chrisjsewell: see above

mmcky:
> > I think that since we're flagging this feature as strictly experimental, and not part of the core myst-notebook spec, we have a little bit of leeway in changing the implementation as we learn more and see how people use it (even it this means removing the implementation entirely!). As long as we make it clear to others what to expect this is IMO ok.
> 
> yes exactly, once it is in and people starting using it, it will be hard to deprecate

I think the main thing is that we introduce and maintain consistent `syntax`. I think what is proposed here is an intuitive way to add a file containing code to a `code-cell`. As with most software we can iterate on the best location on where this is best supported (i.e. promote it to be a more fundamental feature of `myst`) 

````
```{code-cell} python
:file: some-file.py
```
````

chrisjsewell (changes requested):

mmcky:
> my tentativeness here is that file is a pretty generic key

That's what attracts me to `:file:` I would find it easy and natural to remember.

chrisjsewell:
>  I think what is proposed here is an intuitive way to add a file containing code to a code-cell.

The point is that this is not part of standard Jupyter or jupytext, so if you have one of these myst files with this syntax you can't run it as a regular notebook, it has to be converted in a bespoke way first. For example, these files will no longer work with Binder.

I'm going to have to hunt down where we first discussed this lol, because I distinctly recall initially arguing against this bi-directional mapping and you guys were for it.

myst_nb/converter.py
chrisjsewell: https://github.com/executablebooks/MyST-Parser/blob/231971502221d821839c68561ae6293ea8e5382e/myst_parser/docutils_renderer.py#L76

mmcky:
> I'm going to have to hunt down where we first discussed this lol, because I distinctly recall initially arguing against this bi-directional mapping and you guys were for it.

Indeed I am generally all for this and I think it is an important guiding principle. I just think it is ok if we improve support for this `syntax` as project needs demand over time. `Binder` is a good point and possibly a good reason to think about upstreaming this to `jupytext`. However, at this time, we can't really use the `myst` notebooks on binder as there is currently no `jupyter` rendering syntax support for `myst` markup in general which is why we need to use `sphinx-tojupyter` in the short/medium term to produce notebooks that contain base `md`.

choldgraf:
FWIW, I think that @chrisjsewell 's point about `:file:` is a good one. The metadata in `code-cell` becomes generic Jupyter cell metadata, and I could definitely see this stepping on the toes of some other `:file:` metadata.

Two other options:

What about `:include:`? This is:

1. A bit more precise and less common than `file`
2. Already in-line with what Sphinx does

What about `:load:`? This is:

1. Similar to what Jupyter uses for its `%load` magic: https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-load

mmcky:
thanks for these thoughts @choldgraf I think aligning with `jupyter` could be nice using `:load:`? It also works in a similar way to the magic as it needs to be parsed (i.e. executed to load the cell) like `%load`.

mmcky:
thanks @chrisjsewell @choldgraf @AakashGfude for your feedback on this. I think we have arrived at a nice syntax choice and would love to see this merged so we can move forward with our `quantecon` migrations to `jupyter-book`.

AakashGfude (approved):

choldgraf:
I fine w merging this. Though one thought for the long term, could we add an extension point to the myst nb sphinx extension ? Eg if you could register a function that would take a notebook as input, and then hook that up to either a pre or post execute callback, then this kind of functionality could just be a sphinx extension rather than supported natively by myst nb.

We could look at papermill for some inspiration https://papermill.readthedocs.io/en/latest/extending-overview.html

(Or to be honest I wonder if there is some way we could use papermill under the hood for these use cases? That might be more complex tho)

mmcky:
> Eg if you could register a function that would take a notebook as input, and then hook that up to either a pre or post execute callback, then this kind of functionality could just be a sphinx extension rather than supported natively by myst nb.

This is an interesting idea @choldgraf. A `sphinx-extension` could perform actions on any `metadata` associated with a notebook or cell. I will open an issue to track this suggestion.

mmcky:
I thin this now looks good to 🚢 

I have opened this discussion to capture the outstanding `ideas` re: ultimate resting place for this `feature`

https://github.com/executablebooks/meta/discussions/262

jstac:
Thanks for helping with this @choldgraf and @chrisjsewell .  Switching over the main section of our lectures has been stalled on this feature for quite a while and we're a bit stuck right now.  Once it's available we'll be able to push forward with shifting over the rest of the QE material, which is a sizable chunk of what was promised in the grant proposal.

chrisjsewell:
> then this kind of functionality could just be a sphinx extension rather than supported natively by myst nb.

Notebooks should be able to be executed before "entering" sphinx; this was part of the modular design we originally conceived of, i.e. you should be able to use jupyter-cache independently to execute the notebooks and cache them, then run the sphinx build after and it just pulls from this cache.
If you intrisically modify the notebook code cells with sphinx extensions, you can no longer achieve this modularity.

chrisjsewell (changes requested):
In 8621c99 I improved the code, which highlighted an issue that needs fixing: docname is not always available.
If you fix this so that the tests pass, then I will approve

mmcky:
> docname is not available when using jupyter-cache

@chrisjsewell is `docname` not available when using jupyter-cache becuase it is too early in the `sphinx` build process to fill `config.temp_data['docname']` when parsing the document?

chrisjsewell:
> is docname not available when using jupyter-cache becuase it is too early in the sphinx build process to fill config.temp_data['docname'] when parsing the document?

yep, see https://www.sphinx-doc.org/en/master/extdev/appapi.html?highlight=config-inited#sphinx-core-events, docname is only injected in the loop after step 4, execution is handled in step 3

chrisjsewell:
> docname was available before in test_codecell_file_warnings though when issuing the warning which suggests the env is populated at some stage but not at the point that you want to pass env.docname

yeh but this uses the `force` setting not the `cache` setting

mmcky:
> > docname was available before in test_codecell_file_warnings though when issuing the warning which suggests the env is populated at some stage but not at the point that you want to pass env.docname
> 
> yeh but this uses the `force` setting not the `cache` setting

Oh I see. I put 43cc5cc together assuming `docname` wasn't populated for the case of the `master_doc` such as in a single file project context. But I should check this attribute is sensible in a multiple file project context so we aren't always returning the `main` master_doc in the warning.

chrisjsewell:
ok I've basically rewritten the whole thing now lol:

1. I'm not sure what the thinking was behind using a regex to get the filename, given that the options are already passed to a dict. So thats been removed
2. The :load: file is now relative to the markdown file it is in, rather than relative to the sphinx source directory. This is consistent with all other resources (images, downloads, includes, etc)

chrisjsewell:
@AakashGfude I see you are still active, if this looks ok I will merge?

AakashGfude:
> @AakashGfude I see you are still active, if this looks ok I will merge?

@chrisjsewell The code looks ok from my side.👌

choldgraf:
Thanks everybody for all the hard work in thinking this one through and in implementation!

mmcky:
hey @chrisjsewell thanks for the changes. They look good.

akhmerov:
Hi all, nice changes! I'd like to understand the implications of the change to the design principles better. Previously, I thought, that the result of converting myst to ipynb with jupytext, then executing directly, and then converting back should yield the result of what myst-nb does. Is this still the case? If not, what's the design constraint then?

mmcky:
hi @akhmerov I have chosen the current syntax (in part) as it will currently roundtrip (myst:md -> ipynb -> myst:md) through `jupytext` as `metadata` -- so from an `information` perspective it is still compliant -- there is just no action taken on the `:load:` request. 

There has been some discussion around possibly moving this downstream to `jupytext` to action the `:load:` metadata - or possibly convert a `:load:` to `%load` ipython magic -- but that requires some parsing logic which introduces some complexity you may want to avoid in `jupytext`. Currently this enables bringing code in from a file as a `myst-nb` feature (as has therefore been documented in `myst-nb` docs only). 

**Note:** We need this to move the `quantecon lecture` migrations to `jupyter-book` forward as we needed to bring code in from a file. It is becoming a binding constraint and thought this was the "best" part of the stack  to implement support to start with.

choldgraf:
@mmcky it might be helpful if you wrote up the quantecon use case in a discussion or something, so that others know the reason that you need to load one python file across multiple notebooks. Is this for import statements or something?

mmcky:
Sure @choldgraf it is mainly for sharing functions across lectures. You can develop some code in an earlier lecture and then re-use the function or class later (while still displaying the `code`) and only having a single edit point to remove any duplication. 

Once `myst-nb=0.12.0` is released I can certainly at some links to `source` and `output` as a demo from https://github.com/QuantEcon/lecture-python.myst

gibberblot:
This is wonderful. I agree it is useful to be able to replicate code across parts. Does it support importing only part of the file; e.g. a start and end line number? I looked at the code quickly, but it didn't seem so.

choldgraf:
@mmcky thanks for sharing! I think it'd also be helpful to hear why this isn't resolvable by using `import` statements and your own python module (I know that we've discussed before, but I think the writeup could help centralize this information!)

jstac:
I would say it is resolvable in that way but with more effort.

If the code is a bunch of small examples that you don't really want to maintain a package for, and which are frequently edited, then the solution proposed here is much easier to manage.

I suppose the other option is to inject the code into the md files using a preprocessing script before compiling.  I'm not keen on that approach for our source material because our goal is to use plain vanilla JB and its extensions, rather than maintaining extra scripts.

choldgraf:
@jstac just to clarify, I am not arguing against it, I'm just saying that it would be helpful to have a writeup somewhere that clarifies these things, so that we can point to it in discussions, design decisions, etc!

mmcky:
sure thing @choldgraf I will document a use case and design decisions etc. Probably for https://executablebooks.org/en/latest/ if people don't want to distribute this more widely at the moment. I will then link to it from `myst-nb` docs. 

agree with all of @jstac comments. In addition a package is not a good way to show source code -- but is useful for routines that are commonly used. For those items we use the open source [QuantEcon.py](https://github.com/quantecon/QuantEcon.py) package

choldgraf:
@mmcky you could also just make it a "discussion" in the executablebooks/meta discussions board 👍

PR #311, opened 2021-04-01
⬆️ UPDATE: jupyter_sphinx

jakevdp:
Also pinned a few other package versions required for tests to pass.

I want this because `jupyter_sphinx 0.13.2` gets rid of the `"Notebook code has no file extension metadata"` warning that occasionally breaks my builds, and the pip dependency resolver complains if I try to install 0.13.2 with myst-nb.

I know how to work around that warning by adding appropriate metadata to the notebooks, but it's annoying to have to do that for every new notebook, and it's extra boilerplate that shouldn't be necessary.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
         "myst-parser>=0.13.5",
         "docutils>=0.15",
         "sphinx>=2,<4",
-        "jupyter_sphinx>=0.3.2",
+        "jupyter_sphinx~=0.3.2",
         "jupyter-cache~=0.4.1",
         "ipython",
         "nbformat~=5.0",
--- a/setup.py
+++ b/setup.py
@@ -49,11 +49,10 @@ setup(
     python_requires=">=3.6",
     package_data={"myst_nb": ["_static/*"]},
     install_requires=[
-        "myst-parser~=0.13.3",
+        "myst-parser>=0.13.5",
         "docutils>=0.15",
         "sphinx>=2,<4",
-        # TODO 0.3.2 requires some changes to the pytests
-        "jupyter_sphinx==0.3.1",
+        "jupyter_sphinx>=0.3.2",
         "jupyter-cache~=0.4.1",
         "ipython",
         "nbformat~=5.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -49,11 +49,10 @@
     python_requires=">=3.6",
     package_data={"myst_nb": ["_static/*"]},
     install_requires=[
-        "myst-parser~=0.13.3",
+        "myst-parser>=0.13.5",
```
chrisjsewell, line 52:
```suggestion
        "myst-parser~=0.13.5",
```

we should still pin to <0.14

setup.py
```diff
@@ -49,11 +49,10 @@
...
         "docutils>=0.15",
         "sphinx>=2,<4",
-        # TODO 0.3.2 requires some changes to the pytests
-        "jupyter_sphinx==0.3.1",
+        "jupyter_sphinx>=0.3.2",
```
chrisjsewell, line 55:
```suggestion
        "jupyter_sphinx~=0.3.2",
```

we should still pin to <0.4

chrisjsewell (changes requested):
thanks @jakevdp

setup.py
jakevdp: myst-parser v0.13.4 leads to failing tests:
```
============================================================================================================ short test summary info =============================================================================================================
FAILED tests/test_render_outputs.py::test_metadata_image - TypeError: object of type 'NoneType' has no len()
```

setup.py
jakevdp: Sounds good - I thought this was required, but it looks like the issue was elsewhere.

setup.py
chrisjsewell: yep but it still needs to use `~` not `>`

choldgraf: ya I that the `~=` implies "at least as high as this patch release" https://www.python.org/dev/peps/pep-0440/#compatible-release

jakevdp: Oh I see, so your intent here is to exclude any releases `>= 0.14.*`?

choldgraf: Yeah we usually try to pin so that minor version stays the same, but patches can be upgraded. At least until we start using major versions and have proper semver 😅

jakevdp: Got it - thanks!

PR #320, opened 2021-04-25
⬆️ UPGRADE: markdown-it-py v1.0

chrisjsewell:
Updated via dependencies which support this version;

- myst-parser -> 0.14.0
- jupytext -> 1.11.2

See https://github.com/executablebooks/markdown-it-py/blob/master/CHANGELOG.md
and https://github.com/executablebooks/MyST-Parser/blob/master/CHANGELOG.md

Also added mypy type checking for the parser code.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ install_requires =
     ipywidgets>=7.0.0,<8
     jupyter-cache~=0.4.1
     jupyter_sphinx~=0.3.2
-    myst-parser==0.14.0a3
+    myst-parser~=0.14.0
     nbconvert~=5.6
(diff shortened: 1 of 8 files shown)


setup.cfg
```diff
@@ -44,7 +44,7 @@ install_requires =
     ipywidgets>=7.0.0,<8
     jupyter-cache~=0.4.1
     jupyter_sphinx~=0.3.2
-    myst-parser~=0.13.5
+    myst-parser==0.14.0a2
```
chrisjsewell, line 47:
```suggestion
    myst-parser==0.14.0a3
```

PR #334, opened 2021-06-09
Allow Bokeh plots to be glued and pasted

bryanwweber:
Using the JSON-type output from Bokeh, this creates appropriate HTML and
Javascript nodes in the output when the Bokeh+JSON mimetype is found.


bryanwweber:
Whoops, I meant to create this against my own fork. Sorry for the noise!

PR #335, opened 2021-06-09
Allow Bokeh plots to be glued and pasted

bryanwweber:
Using the JSON-type output from Bokeh, this creates appropriate HTML and
Javascript nodes in the output when the Bokeh+JSON mimetype is found.
The Python Bokeh library is only imported when it is needed, to avoid a
package dependency on Bokeh. Users will need to install Bokeh via pip or
conda to be able to glue/paste Bokeh plots. Using the JSON output allows
plots to be pasted into figures as well as a general element with no
further changes.

Uses the Sphinx configuration to decide whether a page has Bokeh content
and only loads BokehJS on those pages. This simplifies the script that
needs to be embedded in each paste and ensures that Bokeh is only loaded
once from the CDN, unless the user runs output_notebook(). In the case
where the user runs output_notebook() the BokehJS library is loaded
twice but this doesn't appear to cause any problems.

### DIFF
--- a/myst_nb/__init__.py
+++ b/myst_nb/__init__.py
@@ -168,6 +168,24 @@ def setup(app: Sphinx):
     app.add_css_file("mystnb.css")
     app.add_domain(NbGlueDomain)
 
+    # Load bokeh
+    def install_bokeh(
+        app: Sphinx, pagename: str, templatename: str, context: "Dict", event_arg: "Any"
+    ) -> None:
+        if app.builder.format != "html":
+            return
+
+        domain = cast(NbGlueDomain, app.env.get_domain("glue"))
+        if domain.has_bokeh(pagename):
+            from bokeh.resources import CDN
+
+            for js_file in CDN.js_files:
+                app.add_js_file(js_file)
+            for js_raw in CDN.js_raw:
+                app.add_js_file(None, body=js_raw, type="text/javascript")  # type: ignore
+
+    app.connect("html-page-context", install_bokeh)
+
     # execution statistics table
     setup_exec_table(app)
 
--- a/myst_nb/nb_glue/__init__.py
+++ b/myst_nb/nb_glue/__init__.py
@@ -1,3 +1,4 @@
+from myst_nb.render_outputs import JB_BOKEH_MIMETYPE
 import IPython
 from IPython.display import display as ipy_display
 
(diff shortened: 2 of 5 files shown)


damianavila:
@choldgraf @chrisjsewell, I am interested to know your thoughts about adding pieces like this one that interfaces with 3rd party libraries. There is some preliminary conversation on the following previous discussion (in case you need some context): https://github.com/executablebooks/meta/discussions/346.

chrisjsewell:
Hey @bryanwweber @damianavila 👋🏼 Essentially I feel these third party interfaces to myst-nb (and thus jupyter-book) should probably be in a separate plugin package, and loaded via entry points

bryanwweber:
Thanks @chrisjsewell Any pointers to how to do that in the docs or elsewhere?

choldgraf:
@bryanwweber there are some docs here:

https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#customise-the-render-process

but this one might be a test-case that drives some enhancements to the documentation (e.g., maybe a short tutorial on implementing your own renderer) I am happy to help write out some steps on how to do this, I think this would be a great way to extend MyST-NB's functionality without adding a lot of complexity to the core (e.g., `myst-nb-bokeh` could be a dependency)

For reference, here's load_renderer function:

https://github.com/executablebooks/MyST-NB/blob/a4c21b16be56725381fd3d829fde6d212d948906/myst_nb/render_outputs.py#L70

and here's where the entry points are defined:

https://github.com/executablebooks/MyST-NB/blob/ce5f69ccce8c11fccaaa9f6d97b3e224154247cf/setup.cfg#L62

bryanwweber:
@choldgraf @chrisjsewell Thanks for the references, I'll look into it! I certainly appreciate the desire to keep the core of MyST-NB small! 😄 

It seems like this plugin would solve the rendering issue, but not the glueing issue. Using the current `myst_nb.glue` on a Bokeh `Figure` stores the HTML for a Bokeh table, plus a plain-text string in the output of the cell. 

I think glueing a Bokeh figure would then require a separate function (e.g., `glue_bokeh()`) defined in the plugin package. To me, it seems like a single UX/API for gluing would be better - why should a user care whether they're gluing a Bokeh figure or Matplolib figure or Python object or anything else?

Do you have any thoughts/opinions about that idea? Thanks!

bryanwweber:
@choldgraf @chrisjsewell OK, so I tried this approach of loading a renderer from an entry point. Unfortunately, this does not allow pasting either, because the renderer is always set to `"default"`: https://github.com/executablebooks/MyST-NB/blob/e39132e4f1c58352a9f6d3004c7336dbbeca4e8a/myst_nb/nb_glue/domain.py#L44

If I change that string to point to my new renderer entry point, then it works :tada: So this is going to definitely require changes to `myst_nb` beyond just potentially importing another extension. I assume you'd like me to make any required changes in a new branch/new PR?

bryanwweber:
@choldgraf @chrisjsewell I'm going to close this. I created an extension package over here: https://github.com/bryanwweber/MyST-NB-Bokeh I'll do some more cleanup and add testing in the next little while. I'll make a separate PR for the fix to MyST-NB that's needed to get it working properly, as I mentioned above ☝️ 

FWIW, the documentation was not super helpful in creating this. I'd be happy to write up a blog post/bit of documentation about how I got all this working. Let me know what's the best place for that. Thanks for the pointers!

chrisjsewell:
Heya yeh no worries cheers

damianavila:
> FWIW, the documentation was not super helpful in creating this. I'd be happy to write up a blog post/bit of documentation about how I got all this working. Let me know what's the best place for that. Thanks for the pointers!

@choldgraf, what would be a good place for this? I do not want to lose the generated momentum created by @bryanwweber 😉

damianavila:
> I'll make a separate PR for the fix to MyST-NB that's needed to get it working properly, as I mentioned above

Btw, that PR lives here (crosslinking for easier ref): https://github.com/executablebooks/MyST-NB/pull/337

PR #336, opened 2021-06-17
📚 DOCS: Update logo

choldgraf:
I tried updating the logo per the latest changes to the MyST logo. It just adds a Helvetica Neue "notebook" (which is the font that Jupyter uses as a secondary font). Couple questions:

- What do others think about it?
- ~What do others think about calling it `myst-notebook` instead of `myst-nb`? Or do folks think it's OK to use them interchangeably?~
    - @damianavila suggested that "notebook" is an overloaded term already, and there's not a strong reason to move from NB to "notebook" given that MyST-NB is already a thing. Makes sense to me, so I'll leave this PR at the logo change and that's it.

current logo:

image

new wide logo:

new square logo:

image


damianavila:
I think I said this in other places, but the notebook word is actually pretty overloaded in the ecosystem.
I would probably keep it as myst-nb (just a thought from someone pretty new in the EBP world, so weigh it accordingly :wink:)

choldgraf:
@damianavila what about something like 

image

that'd be a much smaller departure from the current myst-nb logo, which I think is generally a good thing :-)
?

damianavila:
> what about something like

I think that is definitely better in comparison.

choldgraf:
OK - the latest push updates the logo to use the same structure as our other logos, so it's a pretty minimal change but brings it in-line with our other logos now 👍

damianavila:
LGTM, but since logos are always controversial things, wondering what others think about the change...

choldgraf:
Ya - for reference, I've put the original MyST-NB logo in the top comment. My hope is that this feels more like an "upgrade" (and aligning with our current styles) rather than a "change in the logo"

PR #337, opened 2021-06-22
👌 IMPROVE: Use configured `nb_render_plugin` for glue nodes

bryanwweber:
In the process of creating an extension to render Bokeh outputs, I found that the CellOutputBundleNode was always using the default renderer. The fix is to set the renderer to the user-configured value.

I wasn't sure if a similar change should be made to the inline-type nodes.


bryanwweber:
FYI, my extension is now publicly available on PyPI: https://pypi.org/project/myst-nb-bokeh/ and ReadTheDocs: https://myst-nb-bokeh.readthedocs.io/en/stable/

I'd appreciate a review of this PR so that I don't have to install MyST-NB from a branch in my fork 😄 Thanks!

damianavila:
This seems a sensible request, boosting the signal 😉 
@choldgraf, @chrisjsewell, WDYT?

choldgraf:
It seems correct to make this configuration work like that. So if a plugin is used, is it meant to strictly *extend* the default behavior? Or is that environment variable expected to be set on a per notebook level or something? Maybe I'm not understanding how this particular extension point works well enough

bryanwweber:
> It seems correct to make this configuration work like that. So if a plugin is used, is it meant to strictly *extend* the default behavior? Or is that environment variable expected to be set on a per notebook level or something? Maybe I'm not understanding how this particular extension point works well enough

The way I interpreted the requirement for a plugin is that you have to provide a rendering function for all of the mimetypes that are present in your documents. Whether that comes by extending the built-in rendering class, or writing everything from scratch, is immaterial as long as all the mimetypes are handled. I'm not sure that's correct though. This setting is configured globally in `conf.py`, so there's not (currently) a way to change that from a single notebook, if I understand correctly.

> Also just an FYI the docs failures are because the Myst docs changed, not because of this PR, but we should fix that on master

Thanks, that's what I figured based on the error message but good to have confirmation 😊

chrisjsewell:
Renderers and thus this logic will be significantly changing soon (in https://github.com/executablebooks/MyST-NB/tree/refactor-mystnb).
But let this merge this for now cheers, and I'll loop in you and explain everything when I create the PR for that

PR #344, opened 2021-08-06
🔧 MAINT: Make a more specific selector for no-border

### LINKED ISSUE #287: CSS for cell outputs is too aggressive
This line of the outputs CSS is too aggressive

https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/_static/mystnb.css#L7

The `*` is causing it to strip styling of library-specific outputs that include their own CSS, which changes behavior from the notebook environments. We should choose a less-aggressive CSS rule.

akhmerov:
closes #287 

The original problem was that wildcard CSS leaks, and it's hard to override. Unfortunately, it isn't completely clear to me what the wildcard was supposed to catch in the first place (the description in the source was vague and #287 lacked this info). To figure it out I removed the wildcard and compared the outputs, taking the jupyterbook repo. The proposed change seems to fix the problem with preformatted outputs.

Without wildcard:

image

image

With the proposed change:

image

image

Identical to the original, as far as I can tell. If I overlooked other problems, do let me know.

### DIFF
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -5,7 +5,7 @@ div.container.cell {
 }
 
 /* Removing all background formatting so we can control at the div level */
-.cell_input div.highlight, .cell_input pre, .cell_output .output * {
+.cell_input div.highlight, .cell_input pre, .cell_output pre, .cell_output .output {
   border: none;
   box-shadow: none;
 }
--- a/myst_nb/_static/mystnb.css
+++ b/myst_nb/_static/mystnb.css
@@ -5,7 +5,7 @@ div.container.cell {
 }
 
 /* Removing all background formatting so we can control at the div level */
-.cell_input div.highlight, .cell_input pre, .cell_output pre, .cell_output .output {
+.cell_input div.highlight, .cell_output pre, .cell_input pre, .cell_output .output {
   border: none;
   box-shadow: none;
 }


choldgraf:
Thanks for looking into this - I think that the rest case for this would be using a code cell that generates output with it's own CSS, and confirming that it is not overridden. Did you confirm that as well?

Also I've just enabled precommit ci on this repo so it should fix the jobs when you push another commit

akhmerov:
> Thanks for looking into this - I think that the rest case for this would be using a code cell that generates output with it's own CSS, and confirming that it is not overridden. Did you confirm that as well?

I did. Don't have a screenshot at hand (my book is rebuilding rn and it takes some time), but I am using holoviews, and its UI now renders correctly.

> Also I've just enabled precommit ci on this repo so it should fix the jobs when you push another commit

The errors seem to be in Python, which I didn't touch. Plus precommit seems to pass; are you sure that's the problem?

akhmerov:
I pushed another commit for a good measure.

akhmerov:
Yup, pretty sure the problems aren't mine.

choldgraf:
ah yeah you're right - it looks like the tests are failing on master as well because of a change in how the error messages are in the notebooks. will have to figure that out first, sorry about the noise.

chrisjsewell:
> The errors seem to be in Python, which I didn't touch. 

Yeh I guess its related to https://github.com/executablebooks/jupyter-cache/pull/70; a package (I'm not sure what one) dropped ipykernel as a direct dependency, which in-turn allowed a newer version of ipykernel to be installed.

We should probably update the tests to handle this new error message format, and also maybe pin the ipykernel version in the `testing` extra

chrisjsewell:
The ipykernel now being installed is ipykernel-6.0.3, whereas previously it was ipykernel-5.5.5, so the easiest thing would be to just add `ipykernel~=5.5` to the `testing` extra

akhmerov:
OK, I'll rebase as soon as #347 is merged.

choldgraf:
ok cool - master is fixed and merged into here...is this one ready to go @akhmerov ?

akhmerov:
To the best of my understanding, yes.

As I wrote initially, I can't be fully certain what the wildcard was supposed to catch. The jupyterbook repo itself looks all good, so at least *most* things should work well.

choldgraf:
my intuition is that it's better to have a more restrictive policy that breaks in ways that are easier to debug, rather than a less-restrictive policy that breaks in ways that are hard to debug, so I'd be +1 on just merging this I think

akhmerov:
I completely agree, let's merge! Just don't want to oversell the content :upside_down_face:

PR #347, opened 2021-08-06
🔧 MAINT: Pin ipykernel to ~v5.5

### LINKED ISSUE #346: Tests failing due to regression in ipynb error messages
### Describe the problem

The tests are currently failing because the output syntax for errors seem to have changed:

https://github.com/executablebooks/MyST-NB/runs/3210843582#step:5:233

I tried debugging this locally but it seems like the `ipykernel_NNN` part of the error message changes each time the test is run, so always fails.

@chrisjsewell suggested the following may be the issue:

> The ipykernel now being installed is ipykernel-6.0.3, whereas previously it was ipykernel-5.5.5, so the easiest thing would be to just add `ipykernel~=5.5` to the `testing` extra

choldgraf:
closes #346

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -95,6 +95,7 @@ testing =
     pytest-cov~=2.8
     pytest-regressions
     sympy
+    ipykernel~=5.5
 
 [flake8]
 max-line-length = 100
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,7 @@ rtd =
     altair
     bokeh
     coconut~=1.4.3
+    ipykernel~=5.5
     ipywidgets
     jupytext~=1.11.2
     matplotlib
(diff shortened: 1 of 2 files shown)


chrisjsewell (dismissed):
Cheers! (Providing the tests now pass)

if you could also open an issue for eventually updating the tests to ipykernel v6 (although I think this has dropped support for python 3.6, so we would also need to drop that)

setup.cfg
```diff
@@ -95,6 +95,7 @@ testing =
     pytest-cov~=2.8
     pytest-regressions
     sympy
+    ipykernel~=5.5
```
chrisjsewell, line 98: one nitpick: can we move this up to be in alphabetical order 😬

choldgraf:
thanks for the review - I opened up https://github.com/executablebooks/MyST-NB/issues/348 and moved the dep

choldgraf:
ok gonna merge this one since we're fixing a failing test on master...

PR #352, opened 2021-08-25
MAIN: Enable support for Sphinx>=4

mmcky:
This PR enables support for `sphinx>=4` and drops support for `sphinx<=2`

fixes #338

# Tasks

- [x] rework tests to specify `sphinx3` and `sphinx4` for tests that differ (rather than all test cases)
- [x] Wait for #353 to be merged (which should actually be 0.15.2)
- [x] Rebase onto this branch
- [x] Fix changes needed just for Sphinx 4

This work has all been migrated to #356 so will close this PR. Thanks @chrisjsewell @choldgraf for comments and integration points with `myst-parser`

PR #356, opened 2021-08-30
⬆️ UPGRADE: Upgrade to support sphinx4, remove support for sphinx2

### LINKED ISSUE #338: Allow using Sphinx 4
# Summary

Right now `sphinx>=2,<4` in [setup.cfg](https://github.com/executablebooks/MyST-NB/blob/e39132e4f1c58352a9f6d3004c7336dbbeca4e8a/setup.cfg#L51) makes it impossible to use `myst-nb` with Sphinx 4. `myst-parser` upgraded to Sphinx 4 recently https://github.com/executablebooks/MyST-Parser/pull/390.

# Tasks to complete

- [ ] Try removing the Sphinx to `<5` here
      https://github.com/executablebooks/MyST-NB/blob/master/setup.cfg#L51
      and bump myst-parser to the latest version here:
      https://github.com/executablebooks/MyST-NB/blob/master/setup.cfg#L47
- [ ] Run tests and make changes needed to make them pass
- [ ] Merge in, then we should cut a new release

mmcky:
This PR enables support for `sphinx>=4` and drops support for `sphinx<=2`

fixes #338

Updated `testing` to:

- test `python=3.6` to `python=3.9` against `sphinx3` and `sphinx4` release on `ubuntu-latest`
- test `python=3.9` and `sphinx4` for `osx-latest`
- test `python=3.7`and `sphinx3` for `windows-latest`

This PR adds support for tracking differences between `sphinx3` and `sphinx4` but tracking of `docutils` versions  in the test fixtures is not required at this stage.


mmcky:
@chrisjsewell this PR has far fewer changes than #352 as it turns out there don't seem to be a huge number of `sphinx` specific differences in our `fixture` set for `myst-nb`. In fact only one test in `test_glue.py::test_parser`. 

I have added a string though to `conftests` to enable any future updates to either:

1. what software we need to track in fixtures (i.e. perhaps docutils later on)
2. a single reference point so they all match conventions.

mmcky:
@chrisjsewell @choldgraf I reconfigured `github actions` a little to test all `python` versions for both `sphinx3` and `sphinx4` with the one special case for `windows`.

mmcky:
@chrisjsewell not sure why 

```
tests (ubuntu-latest, 3.6, >=3,<4) Expected — Waiting for status to be reported
tests (ubuntu-latest, 3.8, >=2,<3) Expected — Waiting for status to be reported
```

are expected given the `.github/workflows/tests.yml`

.github/workflows/tests.yml
```diff
@@ -26,12 +26,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
-        sphinx: [">=3,<4"]
+        python-version: [3.7, 3.8, 3.9]
```
chrisjsewell, line 29: If you are deprecating python 3.6 support, then you need to update setup.cfg

chrisjsewell: and also tox.ini

mmcky: ah right -- roger that. 

Perhaps I won't in the case. This can be done separately if we want to do that as a project.

chrisjsewell: the endo of life is in a few months, so it is fine to do so: https://devguide.python.org/#status-of-python-branches

setup.cfg
```diff
@@ -38,7 +38,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    docutils>=0.15
+    docutils>=0.15,<0.17
```
chrisjsewell, line 41: why less than 0.17? this is the latest version and supported by myst-parser

mmcky:
OK so I have updated `testing` to:

- test `python=3.6` to `python=3.9` against the major `sphinx4` release
- test `python=3.8,3.9` against `sphinx3`
- test `python=3.9` and `sphinx4` for `osx-latest`
- test `python=3.7`and `sphinx3` for `windows-latest`

setup.cfg
mmcky: must of misread `myst-parser::setup.cfg` -- I'll update this.

chrisjsewell: (in fact myst-parser specifically tests against 0.17)

chrisjsewell: you may well get changes in the HTML output doing this FYI (if we are testing against those); there were changes from div+class -> semantic HTML tags (hence why myst-parser tests pin docutils to 0.17)

.github/workflows/tests.yml
mmcky: still not sure why

```
tests (ubuntu-latest, 3.8, >=2,<3) Expected — Waiting for status to be reported
```

is expected though.

setup.cfg
mmcky: 👍 I think most of the tests for `myst-nb` are at the `xml` and `ipynb` level. Running the `tox` environment for `py38-sphinx4` installs `docutils=0.17` so not sure we will need to track `docutils` version in fixtures at this stage.

.github/workflows/tests.yml
chrisjsewell: it's part of the branch protection rules

mmcky: @chrisjsewell that end of life chart is pretty great. 👍 I might open an issue to `EOL` python3.6 for `December`

mmcky: > tests (ubuntu-latest, 3.8, >=2,<3) Expected — Waiting for status to be reported

Oh -- thanks I would never have checked there :-)

mmcky:
@chrisjsewell do you want to be explicit about testing of `docutils` versions. In the current `tox` setup:

`docutils=0.16` gets installed for `py##-sphinx3` and
`docutils=0.17` gets installed for `py##-sphinx4`

tests/test_glue/test_parser.sphinx4.xml
```diff
@@ -0,0 +1,169 @@
+<document source="with_glue">
```
choldgraf, line 1: Should we now expect a new one of these files for every major version of Sphinx that we support?

choldgraf (approved):
This in general LGTM! As long as tests are happy then I think we should go for it.

tests/test_glue/test_parser.sphinx4.xml
mmcky: @choldgraf the `sphinx3` is the same as the old fixture file. There is a minor difference in what sphinx includes in the xml "align='default`" is no longer specified in `sphinx4`

mmcky:
@chrisjsewell do you want to cast an eye over this before it is merged?

mmcky:
thanks @choldgraf for reviewing this. I will merge this tomorrow if there are no further comments.

choldgraf:
In the meeting today I think that we agreed that it's time to merge this one and cut a release! 🚀

mmcky:
Absolutely - I can do that this afternoon. Just tied up with a few other things this morning.

PR #364, opened 2021-10-03
✨ NEW: `nb_merge_streams` configuration

chrisjsewell:
If `True`, ensure all stdout / stderr output streams are merged into single outputs.
This ensures deterministic outputs.

### DIFF
--- a/docs/use/start.md
+++ b/docs/use/start.md
@@ -112,4 +112,7 @@ Then for parsing and output rendering:
 * - `nb_output_stderr`
   - `show`
   - One of 'show', 'remove', 'warn', 'error' or 'severe', [see here](use/format/stderr) for details.
+* - `nb_merge_streams`
+  - `False`
+  - If `True`, ensure all stdout / stderr output streams are merged into single outputs. This ensures deterministic outputs.
 `````
--- a/myst_nb/render_outputs.py
+++ b/myst_nb/render_outputs.py
@@ -1,5 +1,6 @@
 """A Sphinx post-transform, to convert notebook outpus to AST nodes."""
 import os
+import re
 from abc import ABC, abstractmethod
 from typing import List, Optional
 from unittest import mock
@@ -91,6 +92,58 @@ def load_renderer(name: str) -> "CellOutputRendererBase":
     raise MystNbEntryPointError(f"No Entry Point found for myst_nb.mime_render:{name}")
 
 
+RGX_CARRIAGERETURN = re.compile(r".*\r(?=[^\n])")
+RGX_BACKSPACE = re.compile(r"[^\n]\b")
+
+
+def coalesce_streams(outputs: List[NotebookNode]) -> List[NotebookNode]:
+    """Merge all stream outputs with shared names into single streams.
+
+    This ensure deterministic outputs.
+
+    Adapted from:
+    https://github.com/computationalmodelling/nbval/blob/master/nbval/plugin.py.
+    """
+    if not outputs:
+        return []
+
+    new_outputs = []
+    streams = {}
+    for output in outputs:
+        if output["output_type"] == "stream":
+            if output["name"] in streams:
+                streams[output["name"]]["text"] += output["text"]
+            else:
+                new_outputs.append(output)
+                streams[output["name"]] = output
+        else:
+            new_outputs.append(output)
+
+    # process \r and \b characters
+    for output in streams.values():
+        old = output["text"]
+        while len(output["text"]) < len(old):
+            old = output["text"]
+            # Cancel out anything-but-newline followed by backspace
+            output["text"] = RGX_BACKSPACE.sub("", output["text"])
+        # Replace all carriage returns not followed by newline
+        output["text"] = RGX_CARRIAGERETURN.sub("", output["text"])
+
+    # We also want to ensure stdout and stderr are always in the same consecutive order,
+    # because they are asynchronous, so order isn't guaranteed.
+    for i, output in enumerate(new_outputs):
+        if output["output_type"] == "stream" and output["name"] == "stderr":
+            if (
+                len(new_outputs) >= i + 2
+                and new_outputs[i + 1]["output_type"] == "stream"
+                and new_outputs[i + 1]["name"] == "stdout"
+            ):
+                stdout = new_outputs.pop(i + 1)
+                new_outputs.insert(i, stdout)
+
+    return new_outputs
+
+
 class CellOutputsToNodes(SphinxPostTransform):
     """Use the builder context to transform a CellOutputNode into Sphinx nodes."""
 
@@ -108,6 +161,8 @@ class CellOutputsToNodes(SphinxPostTransform):
                 renderer_cls = load_renderer(node.renderer)
                 renderers[node.renderer] = renderer_cls
             renderer = renderer_cls(self.document, node, abs_dir)
+            if self.config.nb_merge_streams:
+                node._outputs = coalesce_streams(node.outputs)
             output_nodes = renderer.cell_output_to_nodes(self.env.nb_render_priority)
             node.replace_self(output_nodes)
 
--- a/tests/notebooks/merge_streams.ipynb
+++ b/tests/notebooks/merge_streams.ipynb
@@ -1,0 +1,82 @@
+{
+  "cells": [
+    {
+      "cell_type": "code",
+      "execution_count": 1,
+      "source": [
+        "import sys\n",
+        "print('stdout1', file=sys.stdout)\n",
+        "print('stdout2', file=sys.stdout)\n",
+        "print('stderr1', file=sys.stderr)\n",
+        "print('stderr2', file=sys.stderr)\n",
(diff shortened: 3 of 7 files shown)


docs/use/start.md
```diff
@@ -112,4 +112,7 @@ Then for parsing and output rendering:
 * - `nb_output_stderr`
   - `show`
   - One of 'show', 'remove', 'warn', 'error' or 'severe', [see here](use/format/stderr) for details.
+* - `nb_merge_streams`
+  - `False`
+  - If `True`, ensure all stdout / stderr output streams are merged into single outputs. This ensures deterministic outputs.
```
akhmerov, line 117: This explanation is perhaps worth to clarify further. Which aspect of the outputs becomes deterministic?

chrisjsewell: Too slow lol, new version already out (wanted it for https://sqla-tutorials-nb.readthedocs.io).
Feel free to make a PR with the change though, and I'll merge

akhmerov: I agree, perhaps 5 minutes left to review of a new config option is too short. Maybe something to consider for the future.

Also I can't make this PR exactly because I don't understand what result does merging of the streams achieve. Given that I'm somewhat familiar with nbformat and the involved stack, this is possibly true for many others.

chrisjsewell: The result is "documented" by the test and it output (see PR files changed).
It is a common pattern, already used in places like: https://github.com/computationalmodelling/nbval/blob/master/nbval/plugin.py, and was already planned in https://github.com/executablebooks/jupyter-book/issues/973

akhmerov: Thanks, I understand the purpose of the feature now. I'll trust your judgement call that looking at the source and the test is good enough as far as documentation is concerned.

chrisjsewell: Yeh, I'm certainly open to improved documentation, and we will probably iterate on that more in https://github.com/executablebooks/jupyter-book/pull/1448#issuecomment-933049230

The results, e.g. here https://sqla-tutorials-nb.readthedocs.io/en/latest/metadata.html#emitting-ddl-to-the-database, speak for themselves though: before all the output lines were in separate `div` boxes 😄

mmcky: > The result is "documented" by the test and it output (see PR files changed).
> It is a common pattern, already used in places like: https://github.com/computationalmodelling/nbval/blob/master/nbval/plugin.py, and was already planned in [executablebooks/jupyter-book#973](https://github.com/executablebooks/jupyter-book/issues/973)

Thanks @chrisjsewell -- this is a great improvement 👍 

Thanks @akhmerov for your comments -- I am also not sure what `This ensures deterministic outputs.` means here. 

Just my `2cents`--  I find `documentation` in tests a sub-optimal form of documentation as it requires reading tests, which typically means you need to understand the `test` infrastructure with is non-trivial in many cases.

chrisjsewell: Its deterministic because, instead of having a random (non-deterministic) number of stdout/stderr outputs, you have only a maximum of one for each. As explained in https://github.com/executablebooks/jupyter-book/issues/973

tests/notebooks/merge_streams.ipynb
```diff
@@ -0,0 +1,82 @@
...
+        "import sys\n",
+        "print('stdout1', file=sys.stdout)\n",
+        "print('stdout2', file=sys.stdout)\n",
+        "print('stderr1', file=sys.stderr)\n",
+        "print('stderr2', file=sys.stderr)\n",
+        "print('stdout3', file=sys.stdout)\n",
+        "print('stderr3', file=sys.stderr)\n",
+        "1"
+      ],
+      "outputs": [
```
chrisjsewell, line 16: Here all the stdout/stderr are all mixed up

tests/test_render_outputs/test_merge_streams.xml
```diff
@@ -0,0 +1,23 @@
...
+                import sys
+                print('stdout1', file=sys.stdout)
+                print('stdout2', file=sys.stdout)
+                print('stderr1', file=sys.stderr)
+                print('stderr2', file=sys.stderr)
+                print('stdout3', file=sys.stdout)
+                print('stderr3', file=sys.stderr)
+                1
+        <CellOutputNode classes="cell_output">
+            <literal_block classes="output stream" language="myst-ansi" linenos="False" xml:space="preserve">
```
chrisjsewell, line 14: here they are all in single blocks

mmcky: ah - thanks @chrisjsewell that's helpful. So the order will always be the same (so long as the code-block contents doesn't change).

chrisjsewell: yes the order of stdout / stderr, individually, are guaranteed, so they can be group merged. the order of stdout vs stderr is not guaranteed (they are independent streams), so stderr is placed always after stdout

myst_nb/render_outputs.py
```diff
@@ -91,6 +92,58 @@ def load_renderer(name: str) -> "CellOutputRendererBase":
...
+    # because they are asynchronous, so order isn't guaranteed.
+    for i, output in enumerate(new_outputs):
+        if output["output_type"] == "stream" and output["name"] == "stderr":
+            if (
+                len(new_outputs) >= i + 2
+                and new_outputs[i + 1]["output_type"] == "stream"
+                and new_outputs[i + 1]["name"] == "stdout"
+            ):
+                stdout = new_outputs.pop(i + 1)
+                new_outputs.insert(i, stdout)
```
choldgraf, line 142: So I understand correctly, is this correct:

`new_outputs` may contain a bunch of outputs, some `stdout` and some `stderr`, all interwoven. This section sorts them so that the `stdout` outputs always come after the `stderr` outputs?

and if that is correct, maybe I'm missing something but could we simplify with something like:

```python
sorted(new_outputs, lambda output: output.get("name"))
```

?

docs/use/start.md
choldgraf: I agree that this could be documented more clearly for end-users - if we think that this is a useful feature, and if it is turned off by default, then we should have a section that is easily discoverable for a user that would benefit from this feature.

For example, we could create a parent category of ``## `stdout`/`stderr` `` [around here](https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#removing-stdout-and-stderr), then create a nested section like:

````
### Group `stdout` and `stderr` outputs into a single stream

Notebooks may print multiple things to `stdout` and `stderr`. For example, if a cell prints status updates throughout its execution, each of these is often printed to `stdout`. By default, these outputs may be split across multiple items, and will be rendered as separate "chunks" in your built documentation.

If you'd like each of the outputs in `stderr` and `stdout` to be merged into a single stream for each, use the following configuration:

```
nb_merge_streams = True
```

This will ensure that all `stderr` and `stdout` outputs are merged into a single group.
This also makes the cell outputs more deterministic because 
````

choldgraf (commented):
This is a really nice improvement 👍 . A few specific suggestions for how we can document this better for users.

Also, I agree with others that we should wait for a review before merging this kind of PR, there's no reason to self-merge quickly unless we are fixing a problematic bug or test, or if the PR change is trivial. Getting more 👀 on the PR is always a good idea.

docs/use/start.md
choldgraf: I'm just gonna open a PR to propose these changes since this one is already merged.

myst_nb/render_outputs.py
chrisjsewell: `stdout` come before `stderr`. Again, I would emphasise this is copied directly from https://github.com/computationalmodelling/nbval/blob/master/nbval/plugin.py. If you want to change this, I would suggest making a PR there first

akhmerov: @choldgraf I think it's not what you describe (although I might have misread it).

- non-stream outputs aren't reshuffled. So `[stdout, image, stdout]` stays untouched.
- multiple arguments of the same stream type are merged.

> Again, I would emphasise this is copied directly from https://github.com/computationalmodelling/nbval/blob/master/nbval/plugin.py. If you want to change this, I would suggest making a PR there first

Looking at the description of nbval, I think its purpose is different from publishing notebooks, and therefore the logic it applies might not be optimal for myst-nb or vice versa.

Speaking of publication purposes, reordering streams may be undesirable: imagine demonstrating code that produces a warning halfway through.

On the other hand, coalescing same stream types may be a good idea without even giving the users an option to configure this. Or is there a use case for having two independent text outputs in a row after a single input?

chrisjsewell: >  imagine demonstrating code that produces a warning halfway through.

except, if you write to `sderr`, then there is no guarantee that it will end up halfway `stdout`


anyhow, I have no capacity left to devote to this, so I'll leave you guys to propose new PRs

choldgraf: Good point @akhmerov - I feel like we should see how folks use this feature and see if there are any natural points to improve the ux over time 👍

Issue #367, opened 2021-10-21
Faulty image uri for code-block output images in latex when `--path-output` is specified


AakashGfude:
### Describe the problem

When using `--path-output` cli option in jb with a directory different from the source directory for **latex** build, images generated by code-cells have URI which does not have a proper subdirectory path.

Note that it happens only for projects which have source documents in subdirectories.

The bug was discovered when running latex build in CLI for https://github.com/QuantEcon/lecture-datascience.myst project : https://github.com/QuantEcon/lecture-datascience.myst/runs/3934816817?check_suite_focus=true

### Link to your repository or website

https://github.com/QuantEcon/lecture-datascience.myst

### Steps to reproduce

1. clone https://github.com/QuantEcon/lecture-datascience.myst
2. cd and run `jb build lectures --builder pdflatex --path-output ./ -n --keep-going`
3. images in subdirectories would have path `{../_build/jupyter_execute/plotting_5_1}.png` in tex file, instead of `../_build/jupyter_execute/scientific/plotting_5_1}.png`. Notice that subdirectory is missing from path.

### Additional context

The issue is in https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/render_outputs.py#L467 function, which is responsible for generating image uri.

PR #368, opened 2021-10-21
🐛 FIX: Image URI issue for latex build when source has subdirectories

AakashGfude:
fixes https://github.com/executablebooks/MyST-NB/issues/367 

This PR has one possible fix, splitting with `abspath` pf sphinx_dir to avoid `../`s 
The corresponding issue dives more into the bug details.


mmcky:
@AakashGfude thanks for debugging this issue. Greatly appreciated. 

Should we add a test for this to ensure `images` have the correct path when using  the `--path-output` option?

AakashGfude:
> Should we add a test for this to ensure `images` have the correct path when using the `--path-output` and without using the option?

@mmcky we will have to put this test in jupyter-book then, as --path-output is its feature.

choldgraf:
Is there some way that we can test that the generated paths etc in a Latex build are correct? (either way this seems good to me!)

PR #302, opened 2021-02-28
📚 DOCS: Cleaning up top-level documentation structure

choldgraf:
This cleans up the documentation structure a little bit, and uses parts/captions instead of a single `index` page for some sub-sections.

PR #380, opened 2022-01-08
✨♻️‼️ Rewrite myst-nb

chrisjsewell:
This PR encompasses a top-to-bottom re-write of myst-nb!

It is mainly back compatible, for general users of the extension, with some more technical breaking changes listed below.

----

## Highlights

- docutils API/CLI with command line tools, e.g. `mystnb-docutils-html`
  - Made possible by:
     1. re-working notebook output rendering to be within the initial parse (rather than a post-transform)
     2. moving to use of "native" docutils nodes with attributes
  - Including `glue` role/directives!
- Multi-level configuration (conf.py < notebook level metadata < cell level metadata)
  - Configuration is defined in single class, which is used to generate both docutils and sphinx configuration
  - New config options including: `nb_number_source_lines`, `nb_remove_code_source`, `nb_remove_code_outputs`, `nb_render_error_lexer`
  - See auto-generated config documentation: https://myst-nb--380.org.readthedocs.build/en/380/use/config-reference.html#auto-generated-config
- Parallel friendly (e.g. `sphinx-build -j 4` can execute four notebooks in parallel)
- Page specific loading of ipywidgets JS (sphinx >= 3.5)
- Added raw cell rendering
- Added MIME render plugins
- Better log info/warnings, with `type.subtype` to suppress
- Reworked jupyter-cache integration to be easier to use (including parallel execution)
- New `glue:md` role/directive includes nested parsing of MyST Markdown (see https://myst-nb--380.org.readthedocs.build/en/380/use/glue.html#the-glue-md-role-directive)
- Removes dependencies: jupyter-sphinx, ipywidgets and nbconvert

----

## New API Logic

The API is much clearer/cleaner (see also https://myst-nb--380.org.readthedocs.build/en/380/api/index.html)

The parsing of a notebook consists of a number of stages, with each stage separated into a separate module:

1. `myst_nb.configuration`: The configuration is set (from a file or CLI)
2. `myst_nb.docutils_` / `myst_nb.sphinx_`: The `Parser` is called with an input string and source
3. `myst_nb.read`: The parser reads the input string to a notebook node (possibly using a custom format reader)
4. `myst_nb.execute`: The notebook code outputs are potentially updated, via execution or from a cache
5. `myst_nb.preprocess`: The notebook is "pre-processed" in-place (e.g. to coalesce output streams and extract glue outputs) (this could eventually be pluggable, akin to `nbconvert.preprocessors`)
6. `myst_nb.parse`: The notebook is converted to a Markdown-It tokens syntax tree
7. `myst_nb.render`: The syntax tree is transformed to a docutils document AST (calling the renderer plugin)
8. The docutils document is processed by docutils/sphinx, to create the desired output format(s)

----

## Breaking changes

- A number of configuration option names have been changes, in order for all of them to share the `nb_` prefix.
   - All old names are currently still allowed, and their conversion to the new name automated, with a warning emitted, except...
   - `nb_render_priority` config is replaced by `nb_mime_priority_overrides`, which have different format/semantics
- The AST structure, for code cells etc,  is completely changed.
- ipywidgets needs to be specifically installed, if required for use
- The code cell output render is completely re-written, meaning existing render plugins will no longer work.
- `glue` directives/roles are more restricted on how they work cross-document (only working on the same page by default).
  - `glue:any` directive allows for a `doc` option, and `glue:any`/`glue:text` allow the (relative) doc path to be added before a `::`.


chrisjsewell:
ah just some minor test regression diffs, which should be easy to fix

chrisjsewell:
@choldgraf, @mmcky, @AakashGfude definitely you can initially try installing this branch locally and running the documentation build (e.g. `tox -e docs-clean` and `tox -e doc-update`, potentially after changing something in `conf.py` to trigger a re-build)

chrisjsewell:
Also cc @bryanwweber for the new render plugin (any one else?)
You can look at it in https://github.com/executablebooks/MyST-NB/blob/refactor-mystnb/myst_nb/render.py, but I'll explain more in due course,
and also want to add an example of extending it in https://myst-nb--380.org.readthedocs.build/en/380/use/formatting_outputs.html#customise-the-render-process

chrisjsewell:
Also note I would also we can "revamp" the documentation, but maybe do that in a separate PR (incorporate into #302?)

chrisjsewell:
Ok just diffs failing on Windows, where we use the absolute path for image outputs, rather than a relative one (the expected behaviour) will fix that later

bryanwweber:
Thanks for the heads up @chrisjsewell! I took a quick look just now, and it doesn't seem to be very easy to extend/plug in a new renderer, e.g. my Bokeh one. The reason is because I need to add a new MIME type, which means the render function needs a new `if` branch, which then requires copy-pasting the existing code into my plugin and then extending it.

Ideally, something like the class might grow the ability to register new render functions and the main method loops over the registered functions to render outputs would make for easy extensibility.

chrisjsewell:
Heya,

> Ideally, something like the class might grow the ability to register new render functions and the main method loops over the registered functions to render outputs would make for easy extensibility.

Hmm, yeh registering seemed perhaps a bit overly complex (happy for code suggestions though)
My thinking was, that the easiest way to extend would be to override the `render_unknown` method:

```python
class NewNbElementRenderer(NbElementRenderer):
    def render_unknown(self, data: MimeData) -> List[nodes.Element]:
        if data.mime_type == "cutom mime type":
            return my_custom_nodes
        return super().render_unknown(data)
```

chrisjsewell:
FYI, when I say multi-level configuration, an example would be:

`conf.py`:
```python
nb_execution_mode = "cache"
# this would set the default for every output traceback
nb_render_error_lexer = "ipythontb"
```

`mystnb_page.md`:
````markdown
---
kernelspec: ...
mystnb:  # this is the current default top-level key, but open to discussion
   # this overrides the global configuration for this notebook
   execution_mode: "off"
   render_error_lexer: "other"
---

```{code-cell} ipython3
---
render:
   # override for this cell only
   error_lexer: "none"
---
...
```
````

**Comment by bryanwweber** [comment]
> Heya,
> 
> > Ideally, something like the class might grow the ability to register new render functions and the main method loops over the registered functions to render outputs would make for easy extensibility.
> 
> Hmm, yeh registering seemed perhaps a bit overly complex (happy for code suggestions though) My thinking was, that the easiest way to extend would be to override the `render_unknown` method:
> 
> ```python
> class NewNbElementRenderer(NbElementRenderer):
>     def render_unknown(self, data: MimeData) -> List[nodes.Element]:
>         if data.mime_type == "cutom mime type":
>             return my_custom_nodes
>         return super().render_unknown(data)
> ```
> 
> this is pretty simple @bryanwweber, and doesn't involve any copy/paste?
> 
> (another thought, was that perhaps within the API there would be a method to notify that certain javascript needs to be loaded to HTML pages containing a certain mime type, this is currently semi-hardcoded for only ipywidgets)

Thanks @chrisjsewell! Overriding a function in that way would only permit one extension to be loaded at a time, right? Plus, it's not clear why `render_unknown()` should be overridden, if I know the MIME type, it's not unknown (although this could be documented). In other code, I've seen the use of `importlib.metadata` to find entry points from installed packages, although I can't find a real-world use right now (I'll keep looking, I thought it was in IPython but I can't find it now). The docs are here: https://docs.python.org/3/library/importlib.metadata.html#entry-points It _should_ be reasonably simple to add some code to `__init__` for the renderer to check for entry points, I hope (although I don't know for sure 😜 ) 😄

**Comment by chrisjsewell** [comment]
@bryanwweber the render plugin is already loaded via the `importlib.metadata`

**Comment by bryanwweber** [comment]
> @bryanwweber the render plugin is already loaded via `importlib.metadata` 😉

Ah, awesome! Then do you think it would be possible to store a list/dict of render functions for MIME types and use that to call the appropriate function, rather than the branching structure now?

**Comment by akhmerov** [comment]
Looks nice! How does mimetype priority get taken into account for mime bundles?

File: myst_nb/sphinx_.py [hunk 1]
```diff
@@ -0,0 +1,756 @@
...
+                self.create_warning(
+                    f"Unsupported output type: {output.output_type}",
+                    line=line,
+                    append_to=self.current_node,
+                    wtype=DEFAULT_LOG_TYPE,
+                    subtype="output_type",
+                )
+
+
+class SelectMimeType(SphinxPostTransform):
```
chrisjsewell, line 596: @akhmerov this is where the mime type is selected

chrisjsewell:
> Then do you think it would be possible to store a list/dict of render functions for MIME types and use that to call the appropriate function, rather than the branching structure now?

Well there is different use cases here: you want to handle an unknown (by the default renderer) mime type which is fair. But also others want to override existing mime types, plus also override rendering of stdout/stderr/tracebacks.
So having an entry point for registering new mime types, could possibly be an additional thing, but I still don't think it accounts for all user requirements

myst_nb/sphinx_.py
akhmerov: Thanks. If I understand correctly, this way most of the sphinx machinery sees all the output nodes in the `ast`, until this `post-transform` is encountered. From what I understand so far, this is indeed preferred, and `jupyter-sphinx` uses the same approach.

As a quick note for your consideration: there's an [alternative approach](https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/ast.py#L320-L342) to the post-transform that `jupyter-sphinx` uses. Here, instead of introducing a transform I create a node that presents itself as its highest priority child to the builder. I don't think the differences between the two approaches are important, so I'm mainly bringing this up for completeness.

bryanwweber:
> > Then do you think it would be possible to store a list/dict of render functions for MIME types and use that to call the appropriate function, rather than the branching structure now?
> 
> Well there is different use cases here: you want to handle an unknown (by the default renderer) mime type which is fair. But also others want to override existing mime types, plus also override rendering of stdout/stderr/tracebacks.
> So having an entry point for registering new mime types, could possibly be an additional thing, but I still don't think it accounts for all user requirements

I think if the entry-point loader was able to overwrite objects in the dictionary, it'd be able to satisfy both cases. I'm thinking something like (on my phone so please excuse the pseudo-code)

```
class JPEGRenderer:
    def render(self, data):
        pass  # rendering logic

class Renderer:
    def __init__(self):
        self.renderers = {"image/jpeg": JPEGRenderer,...}
    # load additional renderers from entry points and update the dict
    def render(self, mine, data):
        self.renderers[mime].render(data)
```

Then an extension just has to provide an object that has a `render` method and some way of telling `Renderer` what MIME type it should work for. 

Obviously, this is only one way to approach the problem, and it may be out of scope for this PR. "Good enough" and working software is better than "perfect" and not-working software 😄

chrisjsewell:
Still planning to address the mime type rendering stuff 😉 

But what I've just added, I think is pretty cool: for `text/markdown` outputs, you can now set if you want it rendered as `commonmark` (default, previously hard-coded), `gfm` (https://github.github.com/gfm/), or `myst`.
See:

- https://myst-nb--380.org.readthedocs.build/en/380/use/formatting_outputs.html#markdown
- https://myst-nb--380.org.readthedocs.build/en/380/use/glue.html#the-glue-md-role-directive

myst_nb/sphinx_.py
chrisjsewell: 👍 yep cheers, as you say not too much difference, but I felt the post-transform was slightly more explicit and more of a "standard" sphinx pattern

docs/api/index.rst
```diff
@@ -3,16 +3,70 @@
 Python API
 ==========
 
-.. toctree::
-    :maxdepth: 2
+The parsing of a notebook consists of a number of stages, with each stage separated into a separate module:
 
-    nodes
-    render_outputs
+1. The configuration is set (from a file or CLI)
```
choldgraf, line 8: This is a very helpful description of the whole process, I wonder if we could put this in a place where it is more discoverable? Can do that in a follow up though

docs/use/formatting_outputs.md
```diff
@@ -139,31 +148,67 @@ Image("images/fun-fish.png")
...
 ## Markdown
 
-Markdown output is parsed by MyST-Parser, currently with the configuration set to `myst_commonmark_only=True` (see [MyST configuration options](myst:sphinx/config-options)).
+The format of output `text/markdown` can be specified by `render_markdown_format` configuration:
 
-The parsed Markdown is integrated into the wider documentation, and so it is possible, for example, to include internal references:
+- `commonmark` (default): Restricted to the [CommonMark specification](https://commonmark.org/).
+- `gfm`: Restricted to the [GitHub-flavored markdown](https://github.github.com/gfm/).
+  - Note, this requires the installation of the [linkify-it-py package](https://pypi.org/project/linkify-it-py)
+- `myst`: The MyST parser configuration for the the current document.
```
choldgraf, line 205:
```suggestion
- `myst`: Uses [the MyST parser](https://myst-parser.readthedocs.io/en/latest/) with the same configuration as the current document.
```

docs/use/formatting_outputs.md
```diff
@@ -172,6 +217,49 @@ and even internal images can be rendered!
...
+But setting the `render_markdown_format` to `myst` will allow for more advanced formatting,
+such as including internal references, tables, and even other directives:
+
+`````md
+````{code-cell} ipython3
+---
+render:
+  markdown_format: myst
+---
+display(Markdown('**_some_ markdown** and an [internal reference](use/format/markdown)!'))
```
choldgraf, line 229: InceptionDeeperGIF

docs/use/glue.md
```diff
@@ -316,20 +317,58 @@ Which we reference as Equation {eq}`eq-sym`.
...
+
+- `commonmark` (default): Restricted to the [CommonMark specification](https://commonmark.org/).
+- `gfm`: Restricted to the [GitHub-flavored markdown](https://github.github.com/gfm/).
+  - Note, this requires the installation of the [linkify-it-py package](https://pypi.org/project/linkify-it-py)
+- `myst`: The MyST parser configuration for the the current document.
+
+For example, the following role/directive will glue inline/block MyST Markdown, as if it was part of the original document.
+
+````md
+Here is some {glue:md}`inline_md:myst`!
```
choldgraf, line 352: does this mean people cannot use `:` in their glue keywords? Can we use a less-common separator? e.g. `::` instead of `:`? that would reduce the potential for accidental clashes a good amount

docs/use/glue.md
```diff
@@ -316,20 +317,58 @@ Which we reference as Equation {eq}`eq-sym`.
...
 
-### Pasting from pages you don't include in the documentation
-
-Sometimes you'd like to use variables from notebooks that are not meant to be
-shown to users. In this case, you should bundle the notebook with the rest of your
-content pages, but include `orphan:` in the metadata of the notebook.
-
-For example, the following text: `` {glue:}`orphaned_var` was created in {ref}`orphaned-nb` ``.
-Results in: {glue:}`orphaned_var` was created in {ref}`orphaned-nb`
+### Pasting
```
choldgraf, line 371: I think either this needs to be bumped down to `##` or the subsequent headers should be bumped up to `####`

docs/use/glue.md
```diff
@@ -11,16 +11,17 @@ kernelspec:
   name: python3
 ---
 
-(glue)=
+(glue/main)=
 
 # Insert variables into pages with `glue`
```
choldgraf, line 16: It looks like the ideas around non-Python glue did not make it into this refactor, is that right? Did we hit a hard blocker there? Or is this still a thing to explore in the future?

myst_nb/_static/mystnb.css
```diff
@@ -36,6 +36,7 @@ div.cell_input > div, div.cell_output div.output > div.highlight {
   margin-top: 1em;
 }
 
+/* TODO remove/replace? */
```
choldgraf, line 39: Seems like we can replace this if we are removing the dependency on jupyter_sphinx

choldgraf (commented):
This looks like a nice re-write to me! The structure of the build process seems pretty clean to me, which is great. I made a few comments here and there and happy to keep iterating and providing feedback. A few quick meta thoughts:

- You mentioned that `glue` functionality doesn't work across pages - is your plan to add this back in before we merge this in? Or can you explain the situation with `glue` more generally?
- I'm happy to take a deeper pass at the docs once we merge this in - maybe as part of https://github.com/executablebooks/MyST-NB/pull/302 but potentially just from scratch. @chrisjsewell your list of "what has changed" is quite helpful - if there are any other major user- or developer-facing changes that you can think of, please add them to a laundry list and I'll try to tackle it after this PR.

myst_nb/_static/mystnb.css
chrisjsewell: removed

docs/use/glue.md
chrisjsewell: I just removed that heading entirely

docs/use/glue.md
chrisjsewell: There's a way forward, as a drop in "companion" to Python glue (i.e. you could in principal use them both at the same time, so no deprecations needed), see https://github.com/jupyter/nbclient/pull/160#issuecomment-1005905731
I could essentially write it in to jupyter-cache right now, but obviously prefer it was part of nbclient. Plus ideally we would also introduce a new top-level `user_expressions` key in the nbformat JSON schema

chrisjsewell:
> You mentioned that glue functionality doesn't work across pages - is your plan to add this back in before we merge this in? Or can you explain the situation with glue more generally?

So yes, currently, glue only works for keys within the same page.
This means that it can hook into the same (new) rendering mechanism as standard notebook outputs and, because it is done during the parsing phase, means:

1. You limit any issues of outputs "missing" crucial docutils/sphinx/sphinx-extension transforms/post-transforms (e.g. https://github.com/executablebooks/jupyter-book/issues/1429), i.e. it can be a bit brittle
2. It works both for sphinx and for docutils-only (single page) processing

I can see, though, why it is a shame to lose this functionality, so perhaps we could at least re-introduce it in a more "limited" fashion:

In 1dd2e64, I added code in the sphinx parser, so that glue outputs are now cached (written to file) and the keys, per document name, are stored on the build environment. This means, at a later stage (in a post-transform, you can quickly look if a document has a particular glue key and, if so, load its output.

Not sure for glue roles (where you can't as easily set options), but for glue directives, I could envisage having a `doc` option, e.g.:

````markdown
```{glue:} key
:doc: another/docname
```
````

- If `doc` is set (disallowed for docutils-only parsing), then we create a "pending" node for that directive (rather than using a local key).
- Then, in a post-transform, we look at that document metadata to see if it has the key (rather than requiring keys to be unique over all documents).
- Then, if so, we load the output and run the conversion to docutils nodes.
  - this is the tricky part, because you then have to duplicate "special" conversion code for these conversions and run into issues described above.
  - You could, perhaps, start out limiting it to only the `glue:text` directive/role, because obviously just creating simple `literal` nodes at later stages should not be a problem. It is when you start tying to get fancy and create `image` nodes etc that you run in to potential problems 

---

one final thought

another use mechanism, I could envisage at a later date, and may fit in more with the "live" execution of notebooks model,
is that, instead of only allowing unique keys per notebook, the key you are referencing is its "last state", e.g.

````
```{code-cell}
a = 1
```

{{ a }}

```{code-cell}
a = 2
```

{{ a }}
````

would output `a` as `1` then `2`.

Ideally here, you are not even requiring the user to specify the "glue" keys via the code cell, but perhaps an intermediate that would certainly be possible (see https://github.com/jupyter/nbclient/pull/160#issuecomment-1005905731) is something like:

````
```{code-cell}
---
user_expressions: {"a": "a"}
---
a = 1
```

{glu:}`a`

```{code-cell}
---
user_expressions: {"a": "a"}
---
a = 2
```

{glu:}`a`
````

**Comment by chrisjsewell** [comment]
If you run e.g. `mystnb-docutils-html5 tests/notebooks/complex_outputs.ipynb --nb-execution-mode="off" > test.html`, you will now get an HTML page with all the required CSS loaded 😉

**Comment by mmcky** [comment]
@chrisjsewell I have been building out test branches on the `quantecon` projects. 
They are building nicely in general. I will go through with a closer look later this evening. 

Just a minor question re: command line output

What does `PK=1` mean when notebooks are executed, such as:

```
lecture-python-programming.myst/lectures/about_py.md: Using cached notebook: PK=1 [mystnb]
```

**Comment by chrisjsewell** [comment]
> What does PK=1 mean when notebooks are executed

It means Primary Key, and is the reference for the record stored in the jupyter-cache database, for example you should be able to do:

```
$ pip install jupyter-cache[cli]
$ jcache cache show -p docs/_build/.jupyter_cache 1 
ID: 1
Origin URI: /Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/examples/basic.md
Created: 2022-01-12 00:19
Accessed: 2022-01-12 00:19
Hashkey: b272acfb118246679626f127085bbb07
Data:
  execution_seconds: 3.0643314559999997
```

**Comment by chrisjsewell** [comment]
could maybe change it from `ID` to `PK`

**Comment by mmcky** [comment]
> > What does PK=1 mean when notebooks are executed
> 
> It means Primary Key, and is the reference for the record stored in the jupyter-cache database, for example you should be able to do:
> 
> ```
> $ pip install jupyter-cache[cli]
> $ jcache cache show -p docs/_build/.jupyter_cache 1 
> ID: 1
> Origin URI: /Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/examples/basic.md
> Created: 2022-01-12 00:19
> Accessed: 2022-01-12 00:19
> Hashkey: b272acfb118246679626f127085bbb07
> Data:
>   execution_seconds: 3.0643314559999997
> ```

Oh I see 👍 -- that is a handy reference number. Thats a good example to add to the `docs` too.

**Comment by chrisjsewell** [comment]
> Thats a good example to add to the docs too.

I would certainly like to highlight jupyter-cache more, and the fact that you can execute the notebooks outside of the docs build, etc.
However, one issue at the moment is that it doesn't work well with MyST Markdown or custom format notebooks, because the "conversion function" is not recorded by jupyter-cache.
For example,you can get:

```console
$ jcache stage list
...
nbformat.reader.NotJSONError: Notebook does not appear to be JSON: "---\njupytext:\n  text_representation:\...
```

This is what https://github.com/executablebooks/jupyter-cache/pull/74 will hopefully eventually fix

kai-tub:
Hey, thank you @chrisjsewell for the very exciting work on the myst-nb rewrite.
I am planning on utilizing the custom `myst_nb.read` functionality.
Do have an open to-do list, that I maybe could contribute to?
Thanks! :)

.pre-commit-config.yaml
```diff
@@ -42,17 +42,20 @@ repos:
...
-    hooks:
-    - id: mypy
-      args: [--config-file=setup.cfg]
-      additional_dependencies:
-      - myst-parser~=0.14.0
-      files: >
-        (?x)^(
-            myst_nb/parser.py|
-        )$
+  # - repo: https://github.com/pre-commit/mirrors-mypy
```
choldgraf, line 45: Can you add a comment explaining why this section is commented out, and when we should un-comment it? (or, can we just delete it?)

docs/use/formatting_outputs.md
```diff
@@ -172,6 +217,49 @@ and even internal images can be rendered!
 display(Markdown('![figure](../_static/logo-wide.svg)'))
 ```
 
+But setting the `render_markdown_format` to `myst` will allow for more advanced formatting,
```
choldgraf, line 220: given that this package is explicitly labeled as "myst", why not enable myst markdown parsing by default?

docs/use/formatting_outputs.md
```diff
@@ -247,3 +333,5 @@ You can then select the renderer plugin in your `conf.py`:
 ```python
 nb_render_plugin = "default"
 ```
+
+TODO and example of overriding the renderer ...
```
choldgraf, line 337: We should either add in this example, or make this a comment `% TODO: ...`

myst_nb/read.py
```diff
@@ -0,0 +1,371 @@
...
+        metadata.get("jupytext", {})
+        .get("text_representation", {})
+        .get("format_name", None)
+        != "myst"
+    ):
+        return False
+
+    return True
+
+    # TODO move this to reader, since not strictly part of function objective
```
choldgraf, line 135: Just flagging in case you missed this - unclear if it's TODO in this PR or TODO in the future

choldgraf (commented):
Took another pass and it looks like we are quite close. There are a few quick-ish comments to resolve, but only one complex thing to figure out:

What should we do about cross-page `glue`-ing in this PR? My feeling is that this would be a pretty big breaking change for people if we do not immediately give them an alternative with the same functionality, since cross-page functionality has always been a part of this feature.

At the least, can we define the plan for re-adding this functionality so that if folks aren't happy, we can tell them there's a plan in the works to add the functionality back in?

To see whether this feature is being used in a cross-page way, I searched for `{glue:` using the new github code search (cs.github.com, (which lets you include characters in your search). Here are the first few non-EBP results:

- https://github.com/casact/chainladder-python  <-- ❌ uses `glue` across pages
- https://github.com/popsim-consortium/demes-python  <-- ✔️ uses only on the same page
- https://github.com/FAIRplus/the-fair-cookbook <-- ✔️ uses only on the same page
- https://github.com/cranmer/stats-ds-book <-- ✔️  uses only on same page
- https://github.com/irydium/irydium <-- ❌ documents (and implements) using glue across pages ([this is actually pretty cool](https://github.com/irydium/irydium/blob/a88f6819023f3f579cca0ddee6c9b8f7375bca12/packages/site/static/components/intro.md?q=%7Bglue%3A+language%3AMarkdown#defining-the-code-chunk)).

So I suspect that using `glue` across pages is not the majority of use-cases, but is a significant number.

mmcky:
thanks @choldgraf in identifying some cross-document `glue` cases. @chrisjsewell I was wondering if we could keep the current `glue` behaviour and if a `glu` directive is found by `myst-nb` that can't be resolved it leaves it to be resolved (in a far more limited way) by sphinx?

choldgraf:
### A proposal for mostly keeping `glue` functionality

I was thinking about this a bit more today, and I think your proposal of 

````
```{glue:} key
:doc: another/docname
```
````

Is a nice path forward. I noticed it's similar to how @wlach [implemented glue functionality in irydium](https://irydium.dev/components#introduction).

For roles (or to simplify invocation with directives), we might also support some page metadata like:

```yaml
glue:
  import:
  - path: path/to/stats_ntbk.ipynb
    id: stat
  - path: ../path/to/viz_ntbk.ipynb
    id: viz
```

And then pasting it in could be invoked like:


````
Here's a visualization of the distribution:

```{glue:} viz:myplotkey
```

And the mean is {glu:}`stat:mymeankey` +/- {glu:}`stat:myvarkey`.
````

So there's no "central store" that glue has to know about, and at parse time the paths specified in `doc:` or `imports:` simply must be valid notebook files that can be read with `jupytext` (or perhaps that can be referenced against `jupyter_cache`.

In the future, `key` could also point to something stored in `user_expressions`, or maybe we could even find a way to extend this to cases like "import an entire cell input, cell output, or markdown cell", provided that they had a unique ID attached to them.

Could this give us a path forward on this PR?

also curious if @rowanc1 has thoughts on this API design, given that I think CurveNote has similar kinds of functionality for "refer to pieces of another notebook"

**Comment by rowanc1** [comment]
On the Curvenote side, we reference a particular cell-output of a notebook by a UID, given that cellIds are now part of the notebook standard, it would be great to be able to refer to those directly without doing anything in Python land. (doc=, id=) It would also be fantastic if this were somewhat pluggable, so that you could provide a function to help resolve keys (for example, maybe some of the plots/outputs eventually come from a different project?!).

File: .pre-commit-config.yaml (hunk 9 above)
Comment (chrisjsewell) [reply .pre-commit-config.yaml]: fixed

File: docs/use/formatting_outputs.md (hunk 11 above)
Comment (chrisjsewell) [reply docs/use/formatting_outputs.md]: fixed

**Comment by chrisjsewell** [comment]
Having updated to [jupyter-cache v0.5.0](https://github.com/executablebooks/jupyter-cache/blob/master/CHANGELOG.md#050---2021-01-25), you can now start to see how the integration with myst-nb/jupyter-book would work.
On building the myst-nb docs, part of the stdout is:

```
Using jupyter-cache at: /Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/_build/.jupyter_cache
```

Once the docs are built, we can use the `jcache` CLI to look at this cache:

```console
$ jcache notebook -p /Users/chrisjsewell/Documents/GitHub/MyST-NB-actual/docs/_build/.jupyter_cache list
  ID  URI                               Reader      Added             Status
----  --------------------------------  ----------  ----------------  ---------------
   1  docs/examples/basic.md            myst_nb_md  2022-03-02 04:50  ✅ [1]
   2  docs/examples/coconut-lang.md     myst_nb_md  2022-03-02 04:50  ✅ [2]
   3  docs/examples/custom-formats.Rmd  nbformat    2022-03-02 04:50  ❗️ (unreadable)
   4  docs/examples/interactive.md      myst_nb_md  2022-03-02 04:50  ✅ [4]
   5  docs/use/execute.md               myst_nb_md  2022-03-02 04:50  ✅ [5]
   6  docs/use/formatting_outputs.md    myst_nb_md  2022-03-02 04:50  ✅ [6]
   7  docs/use/glue.md                  myst_nb_md  2022-03-02 04:50  ✅ [7]
   8  docs/use/hiding.md                myst_nb_md  2022-03-02 04:51  ✅ [8]
   9  docs/use/markdown.md              myst_nb_md  2022-03-02 04:51  ✅ [9]
  10  docs/use/orphaned_nb.ipynb        nbformat    2022-03-02 04:51  ✅ [10]
```

You can see that the `.md` files are now stored with a specific reader (entry point), so that jupyter-cache knows how to read them, outside the documentation build.
Reading custom formats is a bit more tricky, I probably won't address that in this PR.

chrisjsewell:
> At the least, can we define the plan for re-adding this functionality so that if folks aren't happy, we can tell them there's a plan in the works to add the functionality back in?

I've opened https://github.com/executablebooks/meta/issues/681, with my detailed thoughts on the topic.
I think we should move the conversation on this topic over to there

> To see whether this feature is being used in a cross-page way, I searched for {glue: using the new github code search
> So I suspect that using glue across pages is not the majority of use-cases, but is a significant number.

Thanks, it's certainly important information to consider, although I don't think we should be completely beholden to current usage, when developing the long-term strategy 😬 Particularly if it is going to be problematic, for any usage of myst outside of sphinx

> I noticed it's similar to how `@wlach` [implemented glue functionality in irydium](https://irydium.dev/components#introduction).

Maybe copy this over to https://github.com/executablebooks/meta/issues/681

>  So there's no "central store" that glue has to know about, at parse time the paths specified in `doc:` or `imports:` simply must be valid notebook files that can be read with jupytext
> Could this give us a path forward on this PR?

This does not seem to address if/when documents have been executed though, for example

- what happens if the `doc` references a document that has yet to be executed?
- what happens if a notebook's code is changed, but there are other documents that reference it via `doc`?

myst_nb/read.py
chrisjsewell: I think this does not need to be strictly addressed in this PR

docs/use/glue.md
chrisjsewell: See https://github.com/executablebooks/meta/issues/681

mmcky:
> > what do you think about "leaving" the unresolvable cases and passing that down into sphinx to resolve
> 
> I'm unclear what you mean by this, given sphinx has no built-in way of handling jupyter noteboook outputs?

I was under the impression that `glue` is currently resolved as a `sphinx.transform` which put limitations around support for various output `mime` types. I guess I was wondering if the parser came across a cross-document `glue` request then it could leave the syntax (as it is currently) and let the sphinx.transform resolve it (with limitations around mime types). But maybe I've misunderstood the current architecture. I don't think is a great solution anyway if the support for `mime types` varies between `in-page` and `cross-page`

chrisjsewell:
> Ideally, something like the class might grow the ability to register new render functions and the main method loops over the registered functions to render outputs would make for easy extensibility.

@bryanwweber see https://myst-nb--380.org.readthedocs.build/en/380/use/formatting_outputs.html#customise-the-render-process

chrisjsewell:
`glue` is now available cross-document, for a small subset of roles/directives/mime types: https://myst-nb--380.org.readthedocs.build/en/380/use/glue.html#pasting-from-other-notebooks

namurphy:
@chrisjsewell — I'm excited that this was merged!  As someone following this project, thank you for doing this!

PR #385, opened 2022-01-20
🧪 TESTS: Pin ipython<8

chrisjsewell:
ipython v8 is only available for Python 3.8+, and it changes exception text, which fails some regression tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -88,6 +88,8 @@ rtd =
 testing =
     coverage<5.0
     ipykernel~=5.5
+    # ipython v8 is only available for Python 3.8+, and it changes exception text
+    ipython<8
     jupytext~=1.11.2
     # TODO: 3.4.0 has some warnings that need to be fixed in the tests.
     matplotlib~=3.3.0


choldgraf:
Huh that minimum python version surprises me - I'd have assumed that `ipython` was following the NumPy/PyData NEP for version pinning: https://numpy.org/neps/nep-0029-deprecation_policy.html

chrisjsewell:
https://github.com/ipython/ipython/blob/46a51ed69cdf41b4333943d9ceeb945c4ede5668/setup.cfg#L29

choldgraf:
ya I saw that - I am just surprised that they decided to do it. I couldn't find an issue where they were discussing the rationale

chrisjsewell:
Oh indeed yeh it is an interesting one

choldgraf:
Ah maybe I got my dates wrong, the commit message where he added this references that NEP as well:

https://github.com/ipython/ipython/commit/5b9aa9c88325012439384b5a704e2b3d183c3366#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7

chrisjsewell:
why does https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table decide to not use https://endoflife.date/python? seems like theres a strange offset

choldgraf:
good question - I am asking around, I wasn't involved in those numpy conversations 🤷

PR #369, opened 2021-10-25
🚀 RELEASE: updates for release v0.13.2

mmcky:
This sets up the release for `0.13.2`

PR #387, opened 2022-01-31
🐛 FIX: Version pinning to execute tests on all supported python versions

AakashGfude:
Since Python 3.6 requires pandas 1.1.x version, and we are supporting Python 3.6 here, so pinning that version.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,7 +78,7 @@ rtd =
     jupytext~=1.11.2
     matplotlib
     numpy
-    pandas~=1.3.5
+    pandas~=1.1.5
     plotly
     sphinx-book-theme~=0.1.0
     sphinx-copybutton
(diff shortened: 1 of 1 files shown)


mmcky:
@AakashGfude the official `python=3.6` deprecation date is: `23 Dec 2021` so I think we should remove `python=3.6` from the testing table now and bump `pandas` up to a more recent version.

chrisjsewell:
This is already done in #380

mmcky:
So let's get tests passing in #369 then we can release `0.13.2`  -- is that OK @chrisjsewell

chrisjsewell:
Bare with me, should be ok but will have a quick think (probably after sleeping)

AakashGfude:
Should we just follow that PR and do the same change here, https://github.com/executablebooks/MyST-NB/pull/380/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR29, and release a version. merging master to that PR then won't introduce any unintended changes.

PR #390, opened 2022-02-06
UPGRADE: sphinx-togglebutton v0.3.0

choldgraf:
This PR does 2 things:

- Raises the pin on `sphinx-togglebutton` to ~= 0.3.0, so that we get the new UI elements that it introduced
- Removes some of the specific instructions in our "hide markdown content" section, because (I believe) we deprecated the ability to hide markdown cells with a cell tag, and so that content was out-of-date. This section now links out to the sphinx-togglebutton and sphinx-design docs, since myst-nb isn't directly responsible for hiding regular ol' content blocks.

### DIFF
--- a/docs/use/hiding.md
+++ b/docs/use/hiding.md
@@ -78,7 +78,8 @@ points =ax.scatter(*data, c=data[0], s=data[0])
 ## Hiding markdown cells
 
 You cannot hide an entire markdown cell, but you can sections of markdown **content** by using roles and directives.
-For more information on how to do this, see [the `sphinx-togglebutton` documentation](https://sphinx-togglebutton.readthedocs.io/en/latest/).
+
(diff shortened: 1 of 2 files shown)


choldgraf:
Hmmm - the regression failures seem related to the myst-nb execution phase. If this is a non-trivial fix then maybe I'll just wait until #380 is in before bumping sphinx-togglebutton, or maybe @chrisjsewell can bump the version as a part of #380 ?

choldgraf:
I'll plan on merging this and cutting a 0.13.2 release, so that we can see these changes in the book theme and update more CSS if need be, unless I hear any objections before then (in particular I don't wanna mess up @chrisjsewell in #380 so lemme know if this would be disruptive)

chrisjsewell:
Yeh I think it's fine cheers

Issue #392, opened 2022-02-11
It should be possible to error and exit the documentation build process if the execution of the notebooks raises an Exception


rdbisme:
### Describe the problem/need and solution

**Context**
We have lots of code cells that execute code that normally should run correctly but when API changes and we forgot to update the docs, the docs build process should failt to force us to fix it. 

**Problem / Idea**
Right now in the `execution.py` errors are just logged, but build continues. Therefore the sphinx build process exit correctly even if the execution of code cells failed. 

**Solution**
If `execution_allow_errors = False`, exceptions should stop the build

**Benefit**
This will avoid to push unreliable code in docs

### Guide for implementation

- I have a PR following.

chrisjsewell:
Hey @rdbisme, I can't remember if this was possible before, but after #380, an execution error will create a `mystnb.exec` type warning, e.g.

```
docs/examples/interactive.md: WARNING: Executing notebook failed: CellExecutionError [mystnb.exec]
docs/examples/interactive.md: WARNING: Notebook exception traceback saved in: docs/_build/html/reports/examples/interactive.err.log [mystnb.exec]
```

In sphinx, to fail on a warning, you can either use `sphinx-build -W` to halt on the first warning (with a non-zero exit code), or `sphinx-build -W --keep-going` to run until the end, but then return a non-zero exit code if any warnings were encountered.
This is exactly how we do the CI in this repo (see also https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning)

You can also use https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings, if you have known warnings that you do not wish to fail on.

I feel that this should address your issue, rather than the non-standard (for sphinx) method of raising an exception proposed in #393?

chrisjsewell:
Actually, sorted it in #404 😉

rdbisme:
Thanks. So you added an option to trigger the Exception? Nice!

PR #393, opened 2022-02-11
Stop documentation build raising Exception if code cells raise Exceptions

rdbisme:
See #392

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ exclude: >
 
 repos:
 
-  - repo: git://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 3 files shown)


chrisjsewell:
Hey @rdbisme thanks for the PR. I am just finishing #380 though, which is a big change of the execution code, so if it's ok, I will look to add this to that

rdbisme:
> Hey @rdbisme thanks for the PR. I am just finishing #380 though, which is a big change of the execution code, so if it's ok, I will look to add this to that

No problem. The only matter of discussion here I think is if this needs an additional config option or `execution_allow_errors` is a good choice.

chrisjsewell:
Closing this for now (see my comment https://github.com/executablebooks/MyST-NB/issues/392#issuecomment-1104951602)
Can always re-open later if required

PR #422, opened 2022-06-08
⬆️ UPGRADE: Add Python 3.10 support

jarrodmillman:


choldgraf (approved):
This LGTM - py310 tests seem happy, and I don't know of any reason to suspect 3.10 wouldn't work, so I'm +1 !

I saw that you deleted the commented-out Sphinx 3 line...In my opinion that's OK since we weren't testing it anyway there, and Sphinx 5 is out now so I suspect we should focus our efforts on supporting that.

chrisjsewell:
cheers

PR #446, opened 2022-09-29
👌 IMPROVE: Replace sphinx-togglebutton with built-in functionality

chrisjsewell:
This allows for tighter integration with myst-nb:

- Nicer rendering of the hidden content
- Customisation of the hide/show prompts

The implementation uses "static" details/summary HTML tags directly + CSS,
similar to [sphinx-design](https://github.com/executablebooks/sphinx-design),
rather than the JS implementation used by sphinx-togglebutton.

### DIFF
--- a/docs/render/hiding.md
+++ b/docs/render/hiding.md
@@ -7,8 +7,7 @@ kernelspec:
 # Hide cell contents
 
 You can use Jupyter Notebook **cell tags** to control some of the behavior of
-the rendered notebook. This uses the [**`sphinx-togglebutton`**](https://sphinx-togglebutton.readthedocs.io/en/latest/)
-package to add a little button that toggles the visibility of content.[^download]
+the rendered notebook.[^download]
 
 [^download]: This notebook can be downloaded as
             **{nb-download}`hiding.ipynb`** and {download}`hiding.md`
@@ -64,6 +63,31 @@ And the following cell has a `hide-cell` tag:
 # This cell has a hide-cell tag
 fig, ax = plt.subplots()
 points =ax.scatter(*data, c=data[0], s=data[0])
+```
+
+You can control the hide/show prompts by using the `code_prompt_show` and `code_prompt_hide` configuration options.
+`{type}` will be replaced with `content`, `source`, or `outputs`, depending on the hide tag.
+See the {ref}`config/intro` section for more details.
+
+````markdown
+
+```{code-cell} ipython3
+:tags: [hide-cell]
+:mystnb:
+:  code_prompt_show: "My show prompt"
+:  code_prompt_hide: "My hide prompt"
+
+print("hallo world")
+```
+````
+
+```{code-cell} ipython3
+:tags: [hide-cell]
+:mystnb:
+:  code_prompt_show: "My show prompt for {type}"
+:  code_prompt_hide: "My hide prompt for {type}"
+
+print("hallo world")
 ```
 
 (use/hiding/markdown)=
--- a/myst_nb/core/config.py
+++ b/myst_nb/core/config.py
@@ -323,6 +323,35 @@ class NbParserConfig:
             ),
         },
     )
+
+    code_prompt_show: str = dc.field(
+        default="Show code cell {type}",
+        metadata={
+            "validator": instance_of(str),
+            "help": "Prompt to expand hidden code cell {content|source|outputs}",
+            "sections": (
+                Section.global_lvl,
+                Section.file_lvl,
+                Section.cell_lvl,
+                Section.render,
+            ),
+        },
+    )
+
+    code_prompt_hide: str = dc.field(
+        default="Hide code cell {type}",
+        metadata={
+            "validator": instance_of(str),
+            "help": "Prompt to collapse hidden code cell {content|source|outputs}",
+            "sections": (
+                Section.global_lvl,
+                Section.file_lvl,
+                Section.cell_lvl,
+                Section.render,
+            ),
+        },
+    )
+
     number_source_lines: bool = dc.field(
         default=False,
         metadata={
--- a/myst_nb/core/render.py
+++ b/myst_nb/core/render.py
@@ -128,38 +128,49 @@ class MditRenderMixin:
     def render_nb_cell_code(self: SelfType, token: SyntaxTreeNode) -> None:
         """Render a notebook code cell."""
         cell_index = token.meta["index"]
+        cell_line = token_line(token, 0) or None
         tags = token.meta["metadata"].get("tags", [])
 
         exec_count, outputs = self._get_nb_code_cell_outputs(token)
 
+        classes = ["cell"]
+        for tag in tags:
+            classes.append(f"tag_{tag.replace(' ', '_')}")
+
         # TODO do we need this -/_ duplication of tag names, or can we deprecate one?
+        hide_cell = "hide-cell" in tags
         remove_input = (
             self.get_cell_level_config(
-                "remove_code_source",
-                token.meta["metadata"],
-                line=token_line(token, 0) or None,
+                "remove_code_source", token.meta["metadata"], line=cell_line
             )
             or ("remove_input" in tags)
             or ("remove-input" in tags)
         )
+        hide_input = "hide-input" in tags
         remove_output = (
             self.get_cell_level_config(
-                "remove_code_outputs",
-                token.meta["metadata"],
-                line=token_line(token, 0) or None,
+                "remove_code_outputs", token.meta["metadata"], line=cell_line
             )
             or ("remove_output" in tags)
             or ("remove-output" in tags)
         )
+        hide_output = "hide-output" in tags
 
         # if we are remove both the input and output, we can skip the cell
         if remove_input and remove_output:
             return
 
+        hide_mode = None
+        if hide_cell:
+            hide_mode = "all"
+        elif hide_input and hide_output:
+            hide_mode = "all"
+        elif hide_input:
+            hide_mode = "input"
+        elif hide_output:
+            hide_mode = "output"
+
         # create a container for all the input/output
-        classes = ["cell"]
-        for tag in tags:
-            classes.append(f"tag_{tag.replace(' ', '_')}")
         cell_container = nodes.container(
             nb_element="cell_code",
             cell_index=cell_index,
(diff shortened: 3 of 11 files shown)


chrisjsewell:
See https://myst-nb--446.org.readthedocs.build/en/446/render/hiding.html#hide-code-cells

choldgraf:
Please give others a chance to review PRs, especially if we are changing UI/UX, and especially if there is not a linked issue that already has discussion and agreement on a change - at least one global working day unless it's something really mission critical to merge quickly.

This seems OK to me - it's pretty nice to be able to configure the text and to have code cell-specific text.

Two design comments:

**sharp edges**: The "rounded corners on top and sharp edges on bottom" UI feels a little bit awkward to me - my vote would have been to just style these like regular buttons rather than have the sharp edges. This feels clunky in particular in the output cells:

image

**a11y**: I suspect that the color contrast of the font isn't strong enough for accessibility / vision-impaired people. Can we increase the contrast of the text?

image

chrisjsewell:
Heya @choldgraf I did indeed need it for an actual live demonstration I gave yesterday 😅 https://aiida-qe-demo.readthedocs.io/en/latest/2_bands_workflow.html

> already has discussion and agreement on a change

The problem is there was no real agreement on sphinx-togglebutton in the first place.
As you know, I wasn't happy with the initial implementation (using the ⊕ buttons), and we discussed offline how to adapt that to essentially use my implementation from sphinx-design (https://github.com/executablebooks/sphinx-togglebutton/pull/32)

This is better, but still when you actually use this in practice, you end up with many

image

which is absolutely not clear to the user what it actually is, until they click it

with the new implementation (post #450) I'm really happy with the UI:
 
image

(expanded)
image

It is now clear, from both the green margin and the prompt, that these are all related to code cells.

For 

image

the goal of the UI is to indicate that this is only the top of the code cell, and that clicking on it will reveal the rest

choldgraf:
I think we can discuss elsewhere, but the point isn't to debate the specific UX of this implementation (the "sharp edges" approach has grown on me and I think it looks nice as well). I think these are the two main points:

- For decisions, we need visibility across the team and community _before a decision has been made_. Ideally each PR should be linked to an issue that provides the context and rationale for needed changes. People (especially non-developers) need to be able to provide feedback about UI/UX that doesn't require coding knowledge. We don't need full consensus to **decide**, but decisions shouldn't be made ad-hoc and unilaterally. Others at least need a chance to observe the rationale and object if they wish.
- For implementations, we need to give time for others on a team to review. This helps spread knowledge about the codebase across multiple people to prevent information silos. It also generally will lead to changes that are broadly acceptable to the team of people working on a repository.

Obviously no project follows both of these perfectly, but we need to begin following practices like this more closely as we grow these projects beyond a small grant-funded team.

To your point above - we _didn't_ follow this process for `sphinx-togglebutton`, and as a result there was disagreement in the end-result that probably could have been smoothed out earlier on. I moved forward quickly there for the same reasons you did here - we needed to iterate and get something out quickly. But we don't want to keep repeating these anti-patterns in the future if we want to grow a community around these tools.

chrisjsewell:
> My point isn't criticize but to start nudging our team towards healthier distributed practices.

yep I completely agree on all of this, and thats why you need to stop doing anything else you're doing on EPB and make your MEP (https://github.com/executablebooks/myst-eps) on the steering council 😉

choldgraf:
Just to be clear, in my opinion we _don't_ want a steering council to make UI/UX decisions about individual repositories in the organization. I think it's important to delegate and distribute our decision-making so that many decisions are made close to the place where the work will happen. The steerco is meant to define strategy, policy, and steward the community. In this case, I think a steering council would decide on the _policy_ for "what process to follow when gathering community feedback for decisions", not on the decision about whether to use sharp edges on code blocks.

chrisjsewell:
I but it defines the top level process of how this is done, and also probably top level goals design decisions for UI/UX. both of which can be refer to when we have these conversations

choldgraf:
### Thank you for this feedback

Kai - I want to express my strong appreciation for this feedback. It is really hard to notice if a project is hard to contribute to, precisely because you don't tend to talk to the people that aren't present. I think one of the most important questions an open project can ask is "who is not in the room right now, and why not?"

Your post provides a number of helpful clues and ideas as to why others may not be able to contribute. As we begin to move the project beyond its initial grant-funded team, I think this will be crucial to improve.

My opinion is that if you, a highly-engaged person with development capabilities, feel this way, then there are many, many more people who feel the same way.  I'm going to provide a few quick responses below - some of them are explanations for points you make but I just want to re-iterate that these aren't making excuses - I recognize that they are all major aspects of growing a community that we must improve.

> So I would like to argue that it is nice (at least for me) if one can follow the decision process of UI changes or any updates at all really.

I think this is crucial. We must separate the discussion and decision-making process from the implementation process. If we don't do this then it will be very hard for non-developers (or really, non maintainers) to participate in these processes.

We have had some discussion of this in the issue below, and I've encoded your suggestions in a comment there there as well. Please provide feedback there if you'd like to see any refinements!

- https://github.com/executablebooks/meta/issues/377#issuecomment-1269578334

> TL;DR: I think every executable books project page should have a community page where external projects can be advertised to minimize duplicated work and let developers feel like they are giving back and not just solving their problems in a public repo.

This is an interesting idea - you think it'd make more sense to delegate this to each sub-project rather than have one centralized place for it?

> TL;DR I personally keep forgetting that the GitHub Discussion Forum exists. Maybe more advertising could help? Or better communicating what can be asked in the forum and that answering those questions is a valuable task?

Agreed - one way that we could resolve this is by using a standardized theme across all of the repositories. This is similar to [what Dask does](https://docs.dask.org/en/stable/). That way they have the same top-level links on any of their documentation pages. Either way, I agree that we should make it clear where to have general discussion in each repository. I've opened an issue to track this here:

- https://github.com/executablebooks/meta/issues/835

> Especially when we start to draw a line between "implementation details" of dependencies like MyST-Parser/Sphinx. Maybe every page could have a toolchain section that "visualizes" the different levels (like visual style of [jupyter book](https://jupyterbook.org/en/stable/explain/components.html)? Maybe videos can help?

Totally agree - at least we could have a centralized place that explained the whole stack and all of its components. This is something that @mmcky has been interested in doing as well, and I know that @rowanc1 likely benefitted from conversation like this as the curvenote team was onboarding as well.

I've encoded this suggestion in this issue, please suggest there if we should refine something:

- https://github.com/executablebooks/meta/issues/836

>  I think there is a need for more 'developer' specific tutorials. Maybe this could be crowd-sourced, but then it should be 'advertised/endorsed' by the executable books forum. But especially the low-level stuff (Sphinx/testing related) is something I think many are missing the expertise on and is something that would best be tackled by the maintainers themselves.

Also agreed here. I believe this is another thing that @mmcky has been interested in for a while now. Here's one issue where he's tracking this:

- https://github.com/executablebooks/meta/issues/672

I think that your suggestion here is slightly more specific, so I've also encoded it in this issue here:

- https://github.com/executablebooks/meta/issues/837

> I would also like know if this is a sign for the move away from Sphinx? As a contributor, I would also like to know where the project is moving in the long-term. But I guess that will be solved with EBP

Good question! Folks on the EBP team are trying to figure this out as well :-). I think that some things are starting to settle into place. Our current goal is to build out the JavaScript implementation of MyST in order to see how difficult it is and what parts of the Jupyter Book stack can be replicated with an implementation that is simpler than Docutils/Sphinx, and integrates more easily with Jupyter. We'd then hope to maintain both the Python and the JavaScript implementations, but define MyST as a community-governed standard that can be re-used across many projects and implementations.

There are some ideas about this at the end of a recent talk I gave on Jupyter Book at the NumFocus conference:

- [Jupyter Book talk](https://docs.google.com/presentation/d/1QXkfnUB-InLD2Dc5BOftJZv-psURTYGYpLn-Nl0ugTw/edit?usp=sharing)

I think what we should do is open up a discussion about some of these ideas in order to invite community feedback...thus far they haven't been crystallized yet but I think they are starting to take shape. You can also preview some of the JavaScript work in these two docs sites:

- https://myst.tools
- https://js.myst.tools

And eventually we will need to figure out how these fit in with the project's "meta" documentatoin:

- https://executablebooks.org

OK I hope that this was some helpful extra context, I am more than happy to discuss with you. I think that making this project more inclusive and adopting healthier asynchronous discussion and development practices is crucial to its success. I want to re-iterate how much I appreciate your thoughtful and helpful comment.

choldgraf:
One other note: I think that many of the challenges that you've described are also in-part because we have not done a good job of growing the initial grant-funded team to include broader participation from the community. This is particularly true for roles like stewarding community forums, triaging and guiding issues, etc. I believe this is one direction that the project should prioritize funding that it gets in the future, but I also think we need to grow our volunteer contributor base in this direction as well. I welcome any suggestions you might have for how we can make it easier for others to contribute directly to the project in whatever form they wish.

kai-tub:
> This [community page] is an interesting idea - you think it'd make more sense to delegate this to each sub-project rather than have one centralized place for it?

> Let's assume that the project is added to a community page on MyST-NB.
I feel like it should belong to this 'documentation' website because it is specific to MyST-NB and doesn't require the full jupyter-book library.
But I assume that most users that would benefit from it would come from the jupyter-book community and not from the MyST-NB only users.
But then the community page would have to be duplicated (especially since this requires changes to the set-up).
As an external contributer, I would be hesitant to ask at two different projects to be included on two community pages.
Maybe there is a way to "inherit" community pages from the lower stack and adding some comments to it?

I thought about this some more, and I think that I don't have a strong preference.
On the one hand, I, as a consumer, would like to have everything in _one place_ to see the ecosystem and popularity of the project + that would be the place I would go to if I wanted to look for alternatives/extensions.
A developer who (potentially) just wants to target MyST-NB and not the jupyter-book community might prefer a different page for each project?
But I guess if somebody targeted MyST-NB, they would be open to also helping JupyterBook users?

The _single_ page just has to be discoverable for users that are looking at the MyST-NB documentation, for example.
At least, that would be my main concern for the single-page solution.

Feel free to move this to a separate issue/discussion.

chrisjsewell:
thanks for the feedback @kai-tub 

> I think I better understand in which direction the project is moving.
> If I contribute to the Python part, how is it shared with the JS implementation? What will be the default?
> Especially the parser and the AST implementation could be provided as a 'golden' standard.

I'd note that I have created https://github.com/executablebooks/myst-eps, precisely for this underpinning design and strategy documentation.
This is exactly the place where such a gold standard would be solidified

Issue #453, opened 2022-10-03
Scroll output horizontally with large Pandas dataframes


dprada:
### Describe the bug

When a pandas dataframe is wider than the output cell, the result is a dataframe printed out beyond the limits of the web when compiled with sphinx. This happens even with the cell tag "output_scroll":

![image](https://user-images.githubusercontent.com/531231/193632478-9a298239-fc73-4a0d-9974-e975b2131f97.png)

I would have expected a horizontal scroll bar as we can see in the notebook opened with jupyter lab:

![image](https://user-images.githubusercontent.com/531231/193632987-d33f1473-2a4a-46f8-98f2-e29d961bc0da.png)

Is the tag "output_scroll" working only vertically?

Thanks in advance for your help with this issue.

### Reproduce the bug

Code to reproduce the behavior:

```python
import pandas as pd
test_dict = {jj*20:[ii] for ii,jj in enumerate('abcdefghijklm')}
pd.DataFrame(test_dict)
```

The sphinx extensions used are:
extensions = [
    'sphinx.ext.autodoc',
    'sphinx.ext.autosummary',
    'sphinx.ext.intersphinx',
    'sphinx.ext.mathjax',
    'sphinx.ext.todo',
    'sphinx.ext.ifconfig',
    'sphinx.ext.viewcode',
    'sphinx.ext.napoleon',
    'sphinx.ext.githubpages',
    'sphinxcontrib.bibtex',
    'sphinx.ext.extlinks',
    'sphinx_remove_toctrees',
    'sphinx_copybutton',
    'myst_nb'
]

### List your environment

Versions used:
```
python: 3.9
sphinx: 5.1.1
myst-nb: 0.16.0
myst-parser: 0.18.0
```

Theme:
html_theme = 'sphinx_rtd_theme'

chrisjsewell:
Heya what theme are you using

dprada:
html_theme = 'sphinx_rtd_theme'

Is this affecting the default behavior?

chrisjsewell:
In sphinx-book-theme I know there is this added CSS: https://github.com/executablebooks/sphinx-book-theme/blob/eb1b6baf098b27605e8f2b7b2979b17ebf1b9540/src/sphinx_book_theme/assets/styles/extensions/_myst-nb.scss#L21
I think that is what you are after

it may be that we want to move this into myst-nb

dprada:
I added those lines to my _static/custom.css file, but it doesn't solve the problem.
[...]

Issue #443, opened 2022-09-14
Add more css to myst-nb code cells


adamgayoso:
### Context

The executable books project has nice css for code cells, but this doesn't carry over when using myst-nb in other themes like Furo.

### Proposal

Add a bit more css to make myst-nb look consistent across themes.

chrisjsewell:
I would be reticent to start trying to accommodate for every theme here, of course people can already overload the CSS with their own via sphinx.
I did already notice though that it wasn't great with furo, particularly for integration with the light/dark switching, which would be good to integrate. At least changing the colors to CSS variables.

I've already worked with pradyunsg to better integrate things like https://github.com/pradyunsg/furo/blob/main/src/furo/assets/styles/extensions/_sphinx-design.sass into furo, so something like this middle-ground would certainly be of interest

chrisjsewell:
Nore, from 0.17.0, the background colors are now specified as CSS variable:

```css
:root {
  --mystnb-source-bg-color: #f7f7f7;
  --mystnb-stdout-bg-color: #fcfcfc;
  --mystnb-stderr-bg-color: #fdd;
  --mystnb-traceback-bg-color: #fcfcfc;
  --mystnb-source-border-color: #ccc;
  --mystnb-source-margin-color: green;
  --mystnb-stdout-border-color: #f7f7f7;
  --mystnb-stderr-border-color: #f7f7f7;
  --mystnb-traceback-border-color: #ffd6d6;
  --mystnb-hide-prompt-opacity: 70%;
  --mystnb-source-border-radius: .4em;
  --mystnb-source-border-width: 1px;
}
```

making it easier to do e.g.

```css
body[data-theme="dark"] {
  --mystnb-source-bg-color: #202020;
  --mystnb-stdout-bg-color: #202020;
  --mystnb-stderr-bg-color: #ee0000;
  --mystnb-traceback-bg-color: #202020;
}
```

eltos:
It would be really nice to have these darkmode definitions mentioned above included, currently it's hard to read the text with e.g. [pydata-sphinx-theme](https://github.com/pydata/pydata-sphinx-theme) with darkmode:
![Screenshot_20221121_192532](https://user-images.githubusercontent.com/19860638/203131768-160b3ab4-871c-4cdd-97ff-a6d108d06776.png)
![Screenshot_20221121_192326](https://user-images.githubusercontent.com/19860638/203131777-081804ea-e964-414b-b624-90304683ddc9.png)

PR #454, opened 2022-10-03
ENH: Adding scroll bars with proper cell tags

dprada:
Some lines were added to mystnb.css to add scroll bars with tags: 'output_scroll', 'scroll-output' and 'scroll-input'.

The lines implemented are similar to the paragraph in https://github.com/executablebooks/sphinx-book-theme/blob/eb1b6baf098b27605e8f2b7b2979b17ebf1b9540/src/sphinx_book_theme/assets/styles/extensions/_myst-nb.scss#L21. The only difference is the inclusion of indications for horizontal bars (width and y lines).

See #453


myst_nb/static/mystnb.css
```diff
@@ -217,6 +217,73 @@ tbody span.pasted-inline img {
...
+	 overflow-x: auto;
+}
+
+div.cell.tag_output_scroll div.cell_output::-webkit-scrollbar, div.cell.tag_scroll-output div.cell_output::-webkit-scrollbar {
+	 width: 0.3rem;
+	 height: 0.3rem;
+}
+
+div.cell.tag_output_scroll div.cell_output::-webkit-scrollbar-thumb, div.cell.tag_scroll-output div.cell_output::-webkit-scrollbar-thumb {
+	 background: #c1c1c1;
```
chrisjsewell, line 240: these colors should be added as CSS variables, e.g. `var(--mystnb-source-border-color)`
have a look at the the top of this file for how they are specified.

also see https://github.com/executablebooks/MyST-NB/issues/443, for why we want to do this; it makes it easier for others to adapt the CSS for integration with their themes

chrisjsewell (changes requested):
thanks @dprada have a look at my comment for changes

myst_nb/static/mystnb.css
dprada: Thanks. That's good practice.

dprada (commented):
This was already solved, right?

bsipocz (approved):
Previous review comments have been addressed, and I've read through the xreffed discussions and think this should be a go ahead.

Thanks @dprada!

Issue #458, opened 2022-11-10
ipywidgets no longer rendering


ClementPinard:
### Describe the bug

**expectation/bug**
the ipywidget should appear in the docs, but doesn't

By looking at the issues history, seems like a recurrent problem, and it is back.

Ipywidgets no longer appear in rendered html pages, and you can actually see it in the official docs :

See it appearing for 0.16
https://myst-nb.readthedocs.io/en/v0.16.0/render/interactive.html#ipywidgets

And see it not appearing for 0.17
https://myst-nb.readthedocs.io/en/v0.16.0/render/interactive.html#ipywidgets

I can confirm for my own docs that 0.17.0 does not show ipywidgets

Interestingly, widgets do appear in the next page about inline code in the official docs, see https://myst-nb.readthedocs.io/en/v0.17.0/render/inline.html#inserting-different-output-types

### Reproduce the bug

just building this repo doc will do I believe.

I also tried to build docs with 0.16 and it didn't work as well, contrary to online docs.

### List your environment

Jupyter Book      : 0.13.1
External ToC      : 0.2.4
MyST-Parser       : 0.18.1
MyST-NB           : 0.17.1
Sphinx Book Theme : 0.3.3
Jupyter-Cache     : 0.5.0
NbClient          : 0.5.13

ClementPinard:
Hi, I suppose you are talking about jupyterbook ?

I didn't use it to build the docs, I used `sphinx-build` instead, using the `conf.py` in the docs folder. Should I do it another way ?

PR #469, opened 2023-01-16
Add ipywidgets javascript

OriolAbril:
Widgets in the docs stopped rendering in https://myst-nb.readthedocs.io/en/latest/render/interactive.html#ipywidgets. Trying out the fix suggested in #458 to see if it fixes the issue.


OriolAbril:
It is working: https://myst-nb--469.org.readthedocs.build/en/469/render/interactive.html#ipywidgets. Not sure if some default might also need to be updated to account for this, or at least add a note in that section with the config that needs to be set

agoose77:
@OriolAbril thank you for this PR! I am going to close it, because I think we need to fix this slightly differently; at the configuration level. 

However, we're very glad for people such as yourself making contributions that fix things like this, and your PR prompted me to do so. Cheers!

agoose77:
Actually, we'll still want this to ensure that the docs widgets work with the latest ipywidgets!

PR #483, opened 2023-04-03
DOCS: fix link to gallery

michaelaye:


agoose77:
Thanks for this! I've confirmed that the new link is valid, so we don't need to wait for CI in this particular case.

agoose77 (approved):

PR #484, opened 2023-04-03
fix: bump `isort` version

agoose77:
`isort` recently released [a bugfix](https://github.com/PyCQA/isort/pull/2078) that fixes Poetry installation. We can upgrade our `isort` version to accommodate this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
       args: [--py37-plus]
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
     - id: isort
 


agoose77:
@chrisjsewell I'd prefer to merge this following your review. Let me know if you're not going to be able to get to it — my judgement here is that this is a non-invasive, well motivated bugfix that only affects our dev environment, so I'm happy to merge it if you don't have availability. Just let me know!

choldgraf:
@agoose77 ah this one got superceded by merging https://github.com/executablebooks/MyST-NB/pull/460 - sorry for the confusion. Generally speaking you should feel free to self-merge clear fixes / maintenance improvements like this as long as they aren't too complex.

PR #485, opened 2023-04-03
ci: use codecov v3 action

agoose77:
Fixes the codecov job in our tests workflow.


choldgraf (approved):
Many thanks @agoose77 :-)

PR #486, opened 2023-04-03
docs: update to latest `sphinx-design`

agoose77:
A cryptic error message during our docs build is fixed in newer versions of `sphinx-design`. This PR just bumps that RTD dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ rtd = [
     "plotly",
     "sphinx-book-theme~=0.3.0",
     "sphinx-copybutton",
-    "sphinx-design~=0.1.0",
+    "sphinx-design~=0.3.0",
     "sphinxcontrib-bibtex",
     "sympy",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ rtd = [
     "plotly",
     "sphinx-book-theme~=0.3.0",
     "sphinx-copybutton",
-    "sphinx-design~=0.3.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -82,7 +82,7 @@ rtd = [
     "plotly",
     "sphinx-book-theme~=0.3.0",
     "sphinx-copybutton",
-    "sphinx-design~=0.1.0",
+    "sphinx-design~=0.3.0",
```
choldgraf, line 85: Wanna bump this to 0.4.1? Or TBH just remove the pin?

agoose77: In this case, because docs are an application rather than a library, we can pin onto `~=0.4.0` or even `==0.4.1`. The former is probably strict enough for now.

agoose77, line 85:
```suggestion
    "sphinx-design~=0.4.0",
```

choldgraf: Yeah i think it's fine to pin too on theoretical grounds, I am just thinking of how to minimize maintainer toil needed against the risk that something randomly breaks (which imo is low in this case), so that we don't need PRs like this in the future

choldgraf (approved):
Lgtm!

pyproject.toml
agoose77: Sure. I can't remember why the docs were broken before this PR, but I suspect it was because we didn't pin something that we should have. I think I'd prefer to manage docs like an app, so that we periodically bump all dependencies, rather than have them break at a random point down the line.

PR #489, opened 2023-04-12
Update quickstart.md to include docs folder

kolibril13:
sphinx projects are often in the "docs folder",
and also the later part of the quick start makes this assumption:
https://myst-nb.readthedocs.io/en/latest/quickstart.html

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -11,7 +11,7 @@ pip install myst-nb
 You can use the `mystnb-quickstart` CLI to quickly create an example Sphinx + MyST-NB project:
 
 ```bash
-mystnb-quickstart my_project/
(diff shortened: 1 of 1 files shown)


kolibril13:
And another idea:
I use virtual python environments, and
`which sphinx-build` points by default to 
`/opt/homebrew/bin/sphinx-build`.
And that causes the error:
```
Extension error:
Could not import extension myst_nb (exception: No module named 'myst_nb')
```
In order to point to the correct sphinx-build in my virtual environment, I had to run 
`pip install sphinx` once, and then it worked without problems

choldgraf (approved):
Seems reasonable to me, thanks !

PR #490, opened 2023-04-12
Add note about how to use cell tags

kolibril13:
I never used cell tags before and after some googling I found this tutorial. Maybe other people might find this link useful as well.


choldgraf (approved):
Thanks!

PR #494, opened 2023-04-15
DOCS: Hint to avoid Extension error in sphinx-build

kolibril13:
I experienced this error, but I found a solution.
This is a hint how to fix it.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -43,7 +43,11 @@ sphinx-build -nW --keep-going -b html docs/ docs/_build/html
 MyST-NB is parallel-friendly, so you can also distribute the build (and execution of notebooks) over *N* processes with: `sphinx-build -j 4`
 :::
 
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -38,10 +38,14 @@ Once you have finished authoring your content, you can now use the [sphinx-build
 sphinx-build -nW --keep-going -b html docs/ docs/_build/html
 ```
 
+
 :::{tip}
 MyST-NB is parallel-friendly, so you can also distribute the build (and execution of notebooks) over *N* processes with: `sphinx-build -j 4`
 :::
 
+In case that you experience the message "Extension error"  in the build log, make sure you call the right sphinx-build command. For python virtual environments, that would be `.venv/bin/sphinx-build  -nW --keep-going -b html docs/ docs/_build/html`.
```
choldgraf, line 46:
````suggestion
```{admonition} The execution environment is the same as your Sphinx environment
Your Sphinx build shares the same environment with the notebooks you execute during a build.
Ensure that you call the correct `sphinx-build` command when building your documentation, or the environment needed to run the notebooks may not be correct.
This often happens if you see an `Extension error`  in the build log, or an error from `jupyter-cache`.
```
````

Issue #310, opened 2021-03-18
Create unique url's / anchors for each code cell


darribas:
**Is your feature request related to a problem? Please describe.**

Very often, once I have created a resource on JB, I find myself wanting to link to specific code cells to illustrate a solution for questions from students, for example.

**Describe the solution you'd like**

I know it is possible to create unique links with the `(<id>)=` pattern but this'd require to manually hardcode it on to every code cell _just in case_/for when some of them will be required.

Ideally, in a similar way to how there is a button that appears on the cell to copy the code, there'd be another one that provides a unique and direct URL to that code cell (it could perhaps be related to the cell number/position in the underlying notebook?).

**Describe alternatives you've considered**

Currently I default to link to the closest section on which that cell belongs to, and _mostly_ works; or create manual links if I know I'll need them.

krassowski:
Anchors for cell based on cell IDs have been implemented in JupyterLab (https://github.com/jupyterlab/jupyterlab/pull/13285) and nbconvert (https://github.com/jupyter/nbconvert/pull/1897). There is also a related issue open on Jupyter Book repo https://github.com/executablebooks/jupyter-book/issues/1812.

krassowski:
Could the `myst_parser` expose the `_heading_slugs` logic from https://github.com/executablebooks/MyST-Parser/pull/718 so that we could add anchors to cells reusing the same reference collection logic?

krassowski:
@chrisjsewell would you agree that the above proposal of exposing `_heading_slugs` logic is going in the right direction? I can open a PR against MyST-parser to complement the preliminary work in https://github.com/executablebooks/MyST-NB/pull/519 but would appreciate an early feedback.

krassowski:
https://github.com/executablebooks/MyST-NB/pull/519 is ready for review; I will need help over there - despite many attempts the slugs do not get embedded as identifiers in the cells. It seems like the target processing code does not get invoked but I cannot narrow it down why.

PR #519, opened 2023-05-21
Add cell IDs

krassowski:
Once finished could fix https://github.com/executablebooks/MyST-NB/issues/310

This an draft to demonstrate changes which may be beneficial upstream.

This PR depends on update of myst-parser to 1.0 as it uses new `_heading_slugs` (depends on https://github.com/executablebooks/MyST-NB/pull/479).


krassowski:
Some tests are failing because random cell id is automatically added when reading a notebook in myst or other text-based formats. The parsers need to be modified to avoid adding cell ids in such case.

PR #531, opened 2023-07-11
Show traceback before raising ExecutionError

paugier:
When

```python
nb_execution_raise_on_error = True
nb_execution_show_tb = True
```

the traceback is not showed, which is very inconvenient when it happens during a CI job!

I guess it is only an issue of incompatibility between these two options, which should be fixed by this swap.

Does it make sense?

### DIFF
--- a/myst_nb/core/execute/cache.py
+++ b/myst_nb/core/execute/cache.py
@@ -80,12 +80,12 @@ class NotebookClientCache(NotebookClientBase):
         # handle success / failure cases
         # TODO do in try/except to be careful (in case of database write errors?
         if result.err is not None:
-            if self.nb_config.execution_raise_on_error:
-                raise ExecutionError(str(self.path)) from result.err
             msg = f"Executing notebook failed: {result.err.__class__.__name__}"
             if self.nb_config.execution_show_tb:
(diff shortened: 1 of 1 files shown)


agoose77:
This PR looks reasonable. For context, I don't *think* this option is that useful. 

- If you're only using `myst-nb`, one can pass the `-T` or `-v` arguments to `sphinx-build`, which show tracebacks in stderr, if `nb_execution_raise_on_error` is set to `True`. 
- If you're using `jupyter-book`, the same applies (enable verbose output `-v`), but it's more tricky to set this option as it's not passed through from `_config.yml`. You can patch the Sphinx config with
  ```yaml
  sphinx:
	recursive_update: true
	config:
	  nb_execution_raise_on_error: true
  ```

Despite this, I don't see any harm in this PR, and it *might* prove useful.

chrisjsewell:
> I don't think this option is that useful ... one can pass the -T or -v arguments to sphinx-build

How are you going to do this when building with ReadTheDocs? https://github.com/executablebooks/MyST-NB/blob/35ebd5429c1c315d2564af261b4df836651962eb/docs/conf.py#L43

PR #537, opened 2023-08-31
Update copyright year to 2023

GlobalMin:
Should be single character change. Not sure what's going on with the last line. If a maintainer could comment and point out what's going on, happy to fix. Might be a blank line removal. Otherwise just updating to 2023.


GlobalMin:
Oh nice, pre-commit to the rescue! 😄

agoose77 (approved):

PR #551, opened 2023-10-13
MAINT: update linting

agoose77:
This PR replaces flake8 with ruff, updates some linting defaults, and fixes mypy.


choldgraf (approved):
This LGTM - I'm not too familiar with ruff but if you think it has a clear benefit over flake8 then let's go for it

PR #552, opened 2023-10-13
ci: test wider matrix

agoose77:
This PR updates our CI to check Python 3.12, and varies `myst-parser`


agoose77:
Let me cut down the number of jobs, slightly.

choldgraf (dismissed):
🚀 - the changes LGTM though note there are some failing tests. I hit approve and leave it to your judgment whether it's right to merge

michaelweinold:
Would 

```
df.a = [r'$\delta$','x','y']
```

instead of the current

```
df.a = [$\delta$','x','y']
```

mitigate the error, @agoose77? Curious that it fails only for some of the tests from the matrix.

michaelweinold:
Relevant line from the readthedocs build:

```
WARNING: The default value for `navigation_with_keys` will change to `False` in the next release.
If you wish to preserve the old behavior for your site, set `navigation_with_keys=True` in the `html_theme_options` dict in your `conf.py` file.
Be aware that `navigation_with_keys = True` has negative accessibility implications:https://github.com/pydata/pydata-sphinx-theme/issues/1492
```

agoose77:
> Would
> 
> ```
> df.a = [r'$\delta$','x','y']
> ```
> 
> instead of the current
> 
> ```
> df.a = [$\delta$','x','y']
> ```
> 
> mitigate the error, @agoose77? Curious that it fails only for some of the tests from the matrix.

This is the fix that I've ultimately chosen, thanks @michaelweinold — I was also thinking of using a non-escaped character. I wasn't sure why docutils fails to escape this properly before I realised that it's being escaped in JSON, not in Python.

michaelweinold:
Well, well, well - the tests look great! Glad I could help with https://github.com/executablebooks/MyST-NB/pull/552#issuecomment-1765101531. Could we get another review by @choldgraf, perhaps? I believe everyone in https://github.com/executablebooks/MyST-NB/issues/543 is eager for a new release.

PR #598, opened 2024-05-06
Update shown code to match source

OriolAbril:
I got very confused checking the shown code and the actual output, then realized the shown code was not synced with the actual source.


agoose77:
Fab! Thank you <3

PR #588, opened 2024-03-11
Fix image metadata

flying-sheep:
Fixes #522

I’m not super interested in getting this to the finish line, so please feel free to take it over and change it in any way you see fit.

The important part is that #522 gets fixed.

The only question is precedence: Will `.get_cell_level_config()` return a width and height if e.g. only a global or notebook-wide default width/height is specified? Which should override which?

PR #599, opened 2024-05-07
Fix image metadata being used in wrong cell

### LINKED ISSUE #605: Always “config changed ('nb_render_image_options')”
### Describe the bug

**context**
When I run sphinx, it  always invalidates the config.

I assume that’s because of a bug that (I introdoced and that) would be fixed in https://github.com/executablebooks/MyST-NB/pull/599

**expectation**
I expected Sphinx’ cache to work

**bug**
But instead Sphinx never realizes that `nb_render_image_options` is unchanges

> updating environment: `[config changed ('nb_render_image_options')]` 359 added, 0 changed, 0 removed

**problem**
This is a problem because local builds should be fast

### Reproduce the bug

Build documentation with `width`/`height` metadata for an image and `myst-nb`

### List your environment

myst-nb==1.1.0

flying-sheep:
Fix bug pointed out in https://github.com/executablebooks/MyST-NB/pull/588#pullrequestreview-2040630500 caused by reusing the image_options dict. Fixes #605.

Seems like in my previous PR I didn’t understand the tests enough to realize that I changed them to nonsense, but now I do.

Sorry for that!

cc @OriolAbril

### DIFF
--- a/tests/notebooks/complex_outputs.ipynb
+++ b/tests/notebooks/complex_outputs.ipynb
@@ -280,6 +280,10 @@
       ]
      },
      "metadata": {
+      "image/png": {
+          "width": 432,
+          "height": 288
+      },
       "needs_background": "light"
      },
      "output_type": "display_data"
--- a/myst_nb/core/render.py
+++ b/myst_nb/core/render.py
@@ -660,7 +660,7 @@ class NbElementRenderer:
         # TODO backwards-compatible re-naming to image_options?
         image_options = self.renderer.get_cell_level_config(
             "render_image_options", data.cell_metadata, line=data.line
-        )
+        ).copy()
         # Overwrite with metadata stored in output
         image_options.update(
             {
--- a/tests/test_parser/test_complex_outputs.xml
+++ b/tests/test_parser/test_complex_outputs.xml
@@ -130,7 +130,7 @@
                 <container classes="cell_output" nb_element="cell_code_output">
                     <container nb_element="mime_bundle">
                         <container mime_type="image/png">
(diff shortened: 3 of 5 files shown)


tests/notebooks/complex_outputs.ipynb
flying-sheep, line 1: Changing the metadata to reflect the `text/plain` representation (`<Figure size 432x288 with 1 Axes>`) results in the plot looking blurry since the image data’s resolution is smaller (390×248).

So while it doesn’t look nice in the test notebook, it makes the most sense like that.

(The perfect solution would be to have quadruple-sized “retina” image data, e.g. 800×600 with the metadata being half that – 400×300 – which would match real-world usage)

juanitorduz:
Hi! It would be great to have this fix into the next release 😄.

agoose77:
Thanks for your feedback everyone! I am doing a round of maintenance on the Sphinx stack at the moment, I will aim for a release next week.

agoose77:
Thanks @flying-sheep, this is very helpful. Another contributor also opened a PR https://github.com/executablebooks/MyST-NB/pull/609, so I'm in the fortunate position of having to make a choice between the two! As there's a new test in #609, I'm going to close this in favour of that one :)

OriolAbril:
@agoose77 I would merge both PRs actually. They do have the addition of the `.copy()` in common, but that is all, so I would not expect merge conflicts either. This PR **fixes** the existing tests ehich are as of now testing that the wrong behaviour happens (consequently, all tests are failing on the other PR). The other PR doesn't change anything about existing tests (which needs to happen too in addition to fixing the bug) and adds a new test that is more specific for image metadata.

If you prefer, I can also merge both PRs into one keeping the commits of each author so contributions are properly tracked and open a new with the changes from both PRs

agoose77:
@OriolAbril yep, I'm pulling the changes from the tests to attribute @flying-sheep :)

PR #609, opened 2024-06-13
FIX: output metadata overwrites image size for all following images

### LINKED ISSUE #608: Static output image size distorts executed output images
### Describe the bug

When you have notebooks that are not executed (`"execution_mode": "off"`) and notebooks that are executed, following images can be distorted.

This happens when the executor that was used for the static notebook has saved the size of embedded images:
```
{
 "data": {
  "image/png": "..."
  "text/plain": [
   "<IPython.core.display.Image object>"
  ]
 },
 "metadata": {
  "image/png": {
   "height": 100,
   "width": 500
  }
 },
 "output_type": "display_data"
}
```

The image size persists onto all following notebooks or cells that are executed by MyST-NB. This does not happen with `jupyter lab` or `nbsphinx`.

### Reproduce the bug

1. Extract the attached [docs.zip](https://github.com/user-attachments/files/15827217/docs.zip) folder or create a `conf.py`:
   ```python
   exclude_patterns = ["_build", "jupyter_execute"]
   extensions = ["myst_nb"]
   ```
   and an `index.md`:
   ````md
   ```{toctree}
   example <example.ipynb>
   ```
   ````
   and the attached notebook file.
4. Build documentation: `sphinx-build -b html docs docs/_build`
5. You might see something like this:
   
   image

### List your environment

(The Github template contains a mistake in the environment command, this project is not `jupyter-book`, but `myst-nb`.)
```shell
$ pip freeze | grep -E "myst-nb|Sphinx"
myst-nb==1.1.0
Sphinx==7.2.6
```

aeisenbarth:
Closes #608

A cell-level configuration parameter was retrieved and then modified. Since it is a dictionary, the global configuration was modified. By making a copy of the dictionary, the overrides to the dictionary are only local to this specific cell.

I considered to modify `get_cell_level_config` to always return a copy, to avoid future mistakes when using this API. But it actually returns not the whole cell config as dictionary, but the value for a key, which can be any type (bool etc., and only for `render_image_options` a dictionary).


OriolAbril:
@agoose77 see https://github.com/executablebooks/MyST-NB/pull/599#issuecomment-2194395623 about the failing tests

agoose77:
This passes all tests, but there are some deployment bugs related to NumPy 2 that we need to address. Merging.

PR #641, opened 2024-10-07
FIX: fix path suffix condition in core/read.py

je-cook:
In #620 the path suffix operation was updated to fix a warning in sphinx 8. The effect was to only check the last suffix instead of all suffixes as was previously achieved. For a filename `my_example.ex.py`  and a custom extension of '.ex.py' the old value of `path.endswith(suffix)` would be true whereas `Path.suffix` only returns '.py' therefore will return false here:

 https://github.com/executablebooks/MyST-NB/blob/83f8d1b9e3499ac7e10ab216fdb946e193fb8407/myst_nb/core/read.py#L71

I've added a test for this but I needed to do quite a bit of playing to get past some of the eccentricities of the test suite. let me know if I need to change anything

The reason for all of this on my end is to distinguish which files I want converted and not all python files in a given directory. For reference see where we do this

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/documentation/source/conf.py#L130-L135


bsipocz (approved):
Thank you @je-cook and I'm sorry for introducing this regression the first place.

PR #657, opened 2025-01-31
add `configuration` key to readthedocs.yml

sneakers-the-rat:
When opening https://github.com/executablebooks/MyST-NB/pull/656, the [build failed](https://readthedocs.org/projects/myst-nb/builds/27027868/) because [an explicit configuration key is now required](https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/)

so anyway here's that

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,5 +14,6 @@ python:
       - requirements: docs/requirements.txt
 
 sphinx:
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
Thank you!

PR #688, opened 2025-07-13
docs: nicer types in docs

henryiii:
There are a couple of warnings locally, let's see what CI shows.


bsipocz (approved):
Thank you so much!

Issue #705, opened 2025-12-20
Support the latest sphinx and python versions 


choldgraf:
Sphinx is about to come out with a new release. We should make sure that our tests and supported versions in metadata support this release as well.

- add test cases for the latest python releases and drop testing for older versions (let's follow the scipy Supported version standards)
- do the same for sphinx
- update our metadata accordingly

PR #706, opened 2025-12-22
Update infrastructure for new pythons and prep for latest sphinx

choldgraf:
This is some prep work for the new version of Sphinx, though doesn't test for it yet because a bunch of other dependencies in the stack still don't support it. So this is mostly some prep work, and the python versions shouldn't be a problem.

- ref #705 
- ref https://github.com/executablebooks/MyST-NB/issues/681

I think the biggest changes are:

- We had to update the matplotlib version, which made us have to update the regression tests that used it.
- We added a compatibility function for new sphinx app behavior

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
     "myst-parser>=1.0.0",
     "nbformat>=5.0",
     "pyyaml",
-    "sphinx>=5",  # If we get bugs for older versions, just pin this higher
+    "sphinx>=5,<9",  # If we get bugs for older versions, just pin this higher
     "typing-extensions",
     # ipykernel is not a requirement of the library,
     # but is a common requirement for users (registers the python3 kernel)


.github/workflows/tests.yml
```diff
@@ -19,32 +19,26 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-        sphinx: [""]  # Newest Sphinx (any)
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        sphinx: ["~=8.0"]
```
bsipocz: -
```suggestion
        sphinx: ["~=8.0"]  # temporary limit, bring it back to newest sphinx once we are sphinx9 compatible
```

tox.ini
```diff
@@ -11,16 +11,14 @@
...
 
 [tox]
-envlist = py311-sphinx7
+envlist = py313-sphinx8
 
 [testenv]
 usedevelop = true
 
-[testenv:py{39,310,311,312,313}-sphinx{5,6,7,8}]
+[testenv:py{310,311,312,313}-sphinx{7,8}]
```
bsipocz: keep the versions that are still supported
```suggestion
[testenv:py{310,311,312,313}-sphinx{5,6,7,8}]
```

bsipocz (dismissed):
This PR is in a better shape than mine, so would merge this first and the do a bit more CI cleanup and dropping of old python.

I have a few comments that I'll commit directly, and will add back an oldest supported job, too. Once CI is all green will go ahead with the merge.

Thank you Chris!

bsipocz (approved):

PR #719, opened 2026-06-28
fix: add webp support

flying-sheep:
I also modified `base_render_priority` to be less repetitive, LMK if if I should modify that part.


bsipocz (approved):
Thank you, this looks reasonable!

PR #721, opened 2026-07-01
feat: allow passing kernel manager to parser

flying-sheep:
Hi, I built [sphinx-exec-jupyter](https://github.com/flying-sheep/sphinx-exec-jupyter#readme) to leverage myst-nb’s machinery for running code blocks in isolation.

In it, I’m using a custom `KernelClient` for efficiency.

This client needs to be passed down all the way to `executenb`, and it would be great if that was possible without monkeypatching `myst-nb`.

## Usage

It can be used directly when using a `Parser` programmaticly:

```py
from myst_nb.sphinx_ import Parser

parser = Parser(km=MyKernelManager())
parser.parse(...)
```

… or by overriding the registered parser in a Sphinx extension like this:

```py
from docutils import nodes
from myst_nb.sphinx_ import Parser

class SomeKernelManager(...): ...

class MyParser(Parser):
    # could override in `__init__` or here
    def parse(self, inputstring: str, document: nodes.document) -> None:
        self.km = MyKernelManager(document.settings.env)
        return super().parse(inputstring, document)

def setup(app):
    app.setup_extension("myst_nb")
    app.add_source_parser(MyParser, override=True)
```

## Possible alternatives

- Instead of adding the `km` property to the parser and passing them to `create_client`, we could wrap `create_client` into a method on `Parser`, then I could override it. The other changes would still be necessary.

- In addition to the above method, I could break up `parse` into more granular steps in general for easier extensibility.

### DIFF
--- a/myst_nb/core/execute/__init__.py
+++ b/myst_nb/core/execute/__init__.py
@@ -10,6 +10,7 @@ from .inline import NotebookClientInline
 
 if TYPE_CHECKING:
     from nbformat import NotebookNode
+    from jupyter_client.manager import KernelManager
 
     from myst_nb.core.config import NbParserConfig
     from myst_nb.core.loggers import LoggerType
@@ -21,6 +22,8 @@ def create_client(
     nb_config: NbParserConfig,
     logger: LoggerType,
     read_fmt: None | dict = None,
+    *,
+    km: KernelManager | None = None,
 ) -> NotebookClientBase:
     """Create a notebook execution client, to update its outputs.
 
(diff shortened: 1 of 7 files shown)


bsipocz:
> Instead of adding the km property to the parser and passing them to create_client, we could wrap create_client into a method on Parser, then I could override it. The other changes would still be necessary.

Could we use a more verbose kwarg/property, e.g. spelling out `kernelmanager` instead of `km`?

flying-sheep:
done!

bsipocz (approved):
Thank you, let's get this in!

PR #725, opened 2026-07-20
fix: switch to kernel_manager_class

flying-sheep:
Sorry for that. Trying to use the new feature, I realized that `jupyter_cache` doesn’t have a way to close the kernel client and neither does nbclient’s `execute`, seems like they don’t really test that API.

This variant works, since it goes down the `owns_kc` path and therefore shuts down everything nicely.

This time I actually checked if it works as expected: https://github.com/flying-sheep/sphinx-exec-jupyter/pull/33

### DIFF
--- a/myst_nb/core/execute/__init__.py
+++ b/myst_nb/core/execute/__init__.py
@@ -10,7 +10,7 @@ from .inline import NotebookClientInline
 
 if TYPE_CHECKING:
     from nbformat import NotebookNode
-    from jupyter_client.manager import KernelManager
+    from jupyter_client import KernelManager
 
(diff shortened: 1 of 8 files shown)


bsipocz:
Thanks @flying-sheep for the follow-up. I'm pretty sure the rtd failure is unrelated but I'll have to have a closer code when I'm back at the keyboard before merging this.

flying-sheep:
Thank you! Again, very sorry, I didn’t consider this.

bsipocz (approved):
RTD failure is unrelated, so let's have this in. Thanks @flying-sheep!

